PR #8, opened 2024-04-03
feat: add useScriptVimeo + Vimeo component

darioferderber:
WIP - showcase for Vimeo component

### DIFF
--- a/src/runtime/components/VimeoEmbed.vue
+++ b/src/runtime/components/VimeoEmbed.vue
@@ -6,12 +6,13 @@
 
 <script lang="ts" setup>
 import { useIntersectionObserver } from "@vueuse/core";
-import { ref, watch, onBeforeUnmount, useId, useScriptVimeo } from "#imports";
+import { ref, onBeforeUnmount, useId, useScriptVimeo } from "#imports";
 
 const props = withDefaults(
   defineProps<{
     videoId: string;
     lazy?: boolean;
+    rootMargin?: string;
     width?: string | number;
     height?: string | number;
     options?: Object;
@@ -21,6 +22,7 @@ const props = withDefaults(
   }>(),
   {
     lazy: true,
+    rootMargin: "50px 50px 50px 50px",
     width: "640",
     height: "360",
     options: () => ({}),
@@ -32,7 +34,7 @@ const props = withDefaults(
 
 // TODO: put events in <script> - after build it doesn't work for some reason
 // error: "both scripts must have same language type" even if they're both written in ts
-const events: Array<string> = [
+const events: string[] = [
   "play",
   "playing",
   "pause",
@@ -85,8 +87,9 @@ const emit = defineEmits([
 ]);
 
 const _id = useId();
-const id = _id.replace('-', '').replace('_', '');
-const status: Ref<string> = ref(null);
+const id = _id.replace("-", "").replace("_", "");
+
+const status: Ref<string | null> = ref(null);
 
 const root = ref(null);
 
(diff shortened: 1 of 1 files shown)


src/runtime/components/Vimeo.vue
```diff
@@ -0,0 +1,191 @@
...
+  play,
+  pause,
+  mute,
+  unmute,
+});
+
+const id = useId();
+const status: Ref<string> = ref(null);
+
+const { Player } = useScriptVimeo();
```
harlan-zw: It would be cool to see the script load in lazily based on whether the component is entering the viewport.

This can be done with the [trigger option](https://unhead.unjs.io/usage/composables/use-script#triggering-script-load).

src/runtime/components/Vimeo.vue
```diff
@@ -0,0 +1,191 @@
+<template>
+  <div :id="id" class="vimeo" :class="{ isPlaying, isLoading, isLoaded }">
+    <button class="vimeo__Play" @click="play" v-if="customPlay && isLoaded">
```
harlan-zw: i'd be leaning towards making this headless and users can copy and paste the slots in as it would be less maintenance/work for end users to customize

src/runtime/registry/vimeo.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+export function useScriptVimeo<T>(options?: VimeoScriptInput, _scriptOptions?: Omit<NuxtUseScriptOptions<T>, 'beforeInit' | 'use'>) {
+  const scriptOptions: NuxtUseScriptOptions<T> = _scriptOptions || {}
+  let Player: VimeoPlayer
+
+  scriptOptions.beforeInit = () => {
+    if (import.meta.client) {
+      Player = function (...params: any[]) {
```
harlan-zw: is there a reason we can't do this within the `use()` function?

src/runtime/registry/vimeo.ts
```diff
@@ -0,0 +1,41 @@
...
+        return new window.Vimeo.Player(...params)
+      }
+    }
+  }
+  
+  return useScript<VimeoScriptApi>({
+    key: 'myCustomScript',
```
harlan-zw: `vimeo`

src/runtime/components/Vimeo.vue
```diff
@@ -0,0 +1,191 @@
...
+  useId,
+  ref,
+  computed,
+  onMounted,
+  onBeforeUnmount,
+  useScriptVimeo,
+} from "#imports";
+
+useHead({
+  title: "Vimeo",
```
harlan-zw: debug artifact?

src/runtime/registry/vimeo.ts
darioferderber: Nope, I guess not, I'll move it to the `use` function.

src/runtime/components/Vimeo.vue
darioferderber:
```vue
<div :id="id" class="vimeo">
  <slot />
</div>
```

something like this?

harlan-zw: Possibly, do we need the class? I haven't used this integration so not too sure what comes out of the box with it

darioferderber: It just creates an `<iframe>` without any class. Maybe it would be better to remove the class attribute so the user can set it's own

src/runtime/components/Vimeo.vue
darioferderber: Great idea, should I add `lazy` prop to the component which is `true` by default?

src/runtime/components/VimeoEmbed.vue
```diff
@@ -0,0 +1,162 @@
...
+
+const root = ref(null);
+
+const { Player, $script } = useScriptVimeo({
+  trigger: props.lazy ? "manual" : undefined,
+});
+
+let player;
+
+if (!props.lazy) $script.then(init);
```
harlan-zw, line 102: really cool :clap: good call on the prop

maybe we could add a prop for the `rootMargin` with a default of ~50px, it should help minimise the CLS

darioferderber: 👍 sure, I'll add that

PR #14, opened 2024-04-10
feat: add usePinterestTag

darioferderber:
Adds Pinterest tag.
I'm not that familiar with Pinterest tag so feel free to suggest changes.

### DIFF
--- a/src/runtime/registry/pinterest-tag.ts
+++ b/src/runtime/registry/pinterest-tag.ts
@@ -13,27 +13,19 @@ export type PinterestEventType =
   | "viewcategory"
   | "custom";
 
+export type PinterestFns = 
+  ((type: "track", eventType: PinterestEventType | string, eventData: object | undefined) => void) &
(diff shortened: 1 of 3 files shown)


src/runtime/registry/pinterest-tag.ts
```diff
@@ -0,0 +1,82 @@
...
+  return useScript<PinterestTagApi>(
+    {
+      key: "pinterestTag",
+      src: "https://s.pinimg.com/ct/core.js",
+      ...options,
+    },
+    {
+      ...scriptOptions,
+      use: () => ({
+        trackPageview: () => window.pintrk("page"),
```
harlan-zw: To reduce the maintenance costs of the third-party scripts we want to avoid adding any of our own logic on top of the APIs exposed by the script itself. So in this instance, we should only return `pintrk`, the end user has the types already so it will be easy enough for them without this abstraction.

darioferderber: Sure, makes sense 👍

PR #53, opened 2024-05-08
chore: add RegistryScript type

huang-julien:
Hey :wave: 

this PR move the content of `RegistryScripts` to `RegistryScript`


harlan-zw:
This looks like a sensible change. For my sake can you try to provide some more context on PRs like this though, just so I can verify if the problem is truly being solved?

Will merge for now though, thanks.

huang-julien:
It's more about DX i'm currently forced to to `RegistryScripts[0]` if i want to type an object before pushing it into the registry

PR #57, opened 2024-05-11
fix: don't use aliases in build-time files

huang-julien:
Currently using build-time types prevent having good strict types in module integrations.

This is due to @nuxt/scripts importing `#nuxt-scripts` alias in type declaration which typescript cannot be aware of, except if the module author extends its tsconfig from a playground tsconfig (but this brings other issues)


harlan-zw:
Seems reasonable, thanks :+1: We can re-export the registry from the runtime if we need runtime type augmentation

PR #58, opened 2024-05-12
feat: expose runtime composables

huang-julien:
following #57 . Module authors that don't extends their tsconfig from the playground canno't import things like `useScripts` from `#imports` without having type issues, we should expose core composables from the package.

Should we improve this by exposing  a single file for `./runtime/core` instead ? or directly `./runtime`

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,14 @@
     "./registry": {
       "types": "./dist/registry.d.ts",
       "import": "./dist/registry.mjs"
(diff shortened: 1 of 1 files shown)


harlan-zw:
If the `#imports` alias is resolved then the `#nuxt-scripts` alias should be resolved, I think there is a better way to solve this rather then requiring explicit imports

PR #61, opened 2024-05-15
fix: fix google namespace augmentation

huang-julien:
Seems like rollup dts doesn't like augmenting namespaces thats are nested with a single `.` like `google.map` on my  side . When building we can get `  namespace must have a "ModuleBlock" body.   ` error

Did you manage to build locally @harlan-zw  ?


harlan-zw:
I have no issues building so this is quite strange, happy to merge though.

PR #65, opened 2024-05-17
fix(vimeo): fix vimeo types

huang-julien:


src/runtime/components/ScriptVimeoPlayer.vue
```diff
@@ -170,7 +178,8 @@ onMounted(() => {
       clickTriggered = false
     }
     for (const event of events) {
-      player!.on(event, (e) => {
+      player!.on(event, (e: EventMap<typeof event>) => {
+        // @ts-expect-error ts isn't able to infer the correct event type
         emits(event, e, player)
```
huang-julien, line 183: not sure how we could correctly infer the type here

src/runtime/registry/vimeo-player.ts
```diff
@@ -4,9 +4,11 @@ import { useRegistryScript } from '../utils'
 import type { RegistryScriptInput } from '#nuxt-scripts'
 import { useHead } from '#imports'
 
+type Constructor<T extends new (...args: any) => any> = T extends new (...args: infer A) => infer R ? new (...args: A) => R : never
+
 export interface VimeoPlayerApi {
   Vimeo: {
-    Player: VimeoPlayerApi
+    Player: Constructor<typeof window.Vimeo>
```
harlan-zw, line 11: :raised_hands:

PR #73, opened 2024-05-24
fix: force useRegistry return type

### LINKED ISSUE #72: `vue-tsc` should pass without errors
Still getting the following type-errors when `@nuxt/scripts` is enabled, even though I don't use these scripts explicitly.

Bildschirmfoto 2024-05-24 um 11 09 46

huang-julien:
Currently building nuxt-scripts makes all registry scripts function type return `any` in declaration files. Forcing the return type of `useRgistryScript` fix this issue

fix #72


harlan-zw:
Thanks!

PR #81, opened 2024-06-04
feat(tpc): build-time third-party-capital composables

huang-julien:
This PR re-introduce TPC as dependency and not type provider. TPC will be used to generate all composables (and components if possible).

Composables and components will be generated at USER build-time and not at @nuxt/scripts.   

steps: 

- [x] review from Aurora team for the implementation
- [x] clean it
- [x] move TPC to peer-deps

optionnal:
- [ ] refactor with magic-ast
- [ ] do it for embed components


src/module.ts
```diff
@@ -120,14 +124,23 @@ export default defineNuxtModule<ModuleOptions>({
...
+    })
+
     addComponentsDir({
       path: resolve('./runtime/components'),
     })
 
     const scripts = registry(resolve)
+    scripts.push({
+      ...(await addGoogleTagManager()).registry,
+      ...(await addGoogleAnalytics()).registry,
```
flashdesignory: should we check if an entry with the same id exists and replace it? 
Seems safer, to avoid dupes (not sure how it could happen, but maybe good to safe-guard)

src/registry.ts
```diff
@@ -4,28 +4,29 @@ import type { IntercomInput } from './runtime/registry/intercom'
...
-      }),
-      import: {
-        name: 'useScriptGoogleAnalytics',
-        from: resolve('./runtime/registry/google-analytics'),
-      },
-    },
+    // {
+    //   label: 'Google Analytics',
+    //   category: 'analytics',
+    //   logo: `<svg xmlns="http://www.w3.org/2000/svg" width="28.85" height="32" viewBox="0 0 256 284"><path fill="#F9AB00" d="M256.003 247.933a35.224 35.224 0 0 1-39.376 35.161c-18.044-2.67-31.266-18.371-30.826-36.606V36.845C185.365 18.591 198.62 2.881 216.687.24a35.221 35.221 0 0 1 39.316 35.16z"/><path fill="#E37400" d="M35.101 213.193c19.386 0 35.101 15.716 35.101 35.101c0 19.386-15.715 35.101-35.101 35.101S0 267.68 0 248.295c0-19.386 15.715-35.102 35.101-35.102m92.358-106.387c-19.477 1.068-34.59 17.406-34.137 36.908v94.285c0 25.588 11.259 41.122 27.755 44.433a35.161 35.161 0 0 0 42.146-34.56V142.089a35.222 35.222 0 0 0-35.764-35.282"/></svg>`,
```
flashdesignory: general question: what is the svg logo used for? 
It doesn't have any functional purpose that I can think of.

src/tpc/google-analytics.ts
```diff
@@ -0,0 +1,54 @@
...
+    filename,
+    registry: {
+      label: 'Google Tag Manager',
+      category: 'tracking',
+      logo: `<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 256 256"><path fill="#8AB4F8" d="m150.262 245.516l-44.437-43.331l95.433-97.454l46.007 45.091z"/><path fill="#4285F4" d="M150.45 53.938L106.176 8.731L9.36 104.629c-12.48 12.48-12.48 32.713 0 45.207l95.36 95.986l45.09-42.182l-72.654-76.407z"/><path fill="#8AB4F8" d="m246.625 105.37l-96-96c-12.494-12.494-32.756-12.494-45.25 0c-12.495 12.495-12.495 32.757 0 45.252l96 96c12.494 12.494 32.756 12.494 45.25 0c12.495-12.495 12.495-32.757 0-45.251"/><circle cx="127.265" cy="224.731" r="31.273" fill="#246FDB"/></svg>`,
+      import: {
+        name: 'useScriptGoogleAnalytics',
+        from: dst,
+      },
+      // todo can probably do better. Change or expose utils in tpc ?
```
flashdesignory: @huang-julien - do you need anything exposed from the tpc package or from the local tpc folder?

flashdesignory (commented):
This looks really interesting. There's definitely some Nuxt specific stuff that I am not too familiar with, but overall it makes sense. 

I like it, assuming we're ok with the manual / semi-manual updates of the tpc package. 

If I'm getting it right, the gist is that we're adding (or updating) tpc scripts in the registry and kinda "glueing it all together", right?

src/registry.ts
harlan-zw: Nuxt Devtools and the docs, it shouldn't get bundled into runtime code

src/tpc/google-analytics.ts
```diff
@@ -0,0 +1,54 @@
...
+import { GoogleAnalytics, GoogleTagManager } from 'third-party-capital'
+import type { RegistryScript } from '../runtime/types'
+import { getTpcScriptContent } from './utils'
+
+export default async function addGoogleAnalytics() {
+  const { dst, filename } = addTemplate({
+    write: true,
+    getContents() {
+      return getTpcScriptContent({
+        data: GoogleAnalytics({}),
```
flashdesignory: since we're not passing in any values here... would it be useful for tpc to export the raw JSON files?
we could also expose the formatData function if that's helpful in some way.

src/module.ts
huang-julien: Yeah, it's just while writing the PR. it's messy right now, i'll clean it before submiting it

src/tpc/google-analytics.ts
huang-julien: Yeah I was thinking about asking that

src/tpc/utils.ts
```diff
@@ -0,0 +1,118 @@
...
+            src: withQuery('${mainScript.url}', {${mainScript.params?.map(p => `${p}: options?.${p}`)}})
+        },
+        ${nuxt.options.dev ? 'schema: OptionSchema,' : ''}
+        scriptOptions: {
+            use: ${input.use.toString()},
+            stub: import.meta.client ? undefined :  ${input.stub.toString()}
+        },
+        ${clientInitCode.length ? `clientInit: import.meta.server ? undefined : () => {${clientInitCode.join('\n')}},` : ''}
+    }), _options)
+}`)
```
huang-julien: Moving to magicAST or AST is probably not worth and too complex

flashdesignory (approved):
Approving implementation approach from Aurora side - nice work!

flashdesignory:
would love a follow up with your comment: "We can also add a version check from npm later."
That'll be great to show a message to the user, if they're using an outdated version of tpc

src/registry.ts
```diff
@@ -4,28 +4,29 @@ import type { IntercomInput } from './runtime/registry/intercom'
...
-      logo: `<svg xmlns="http://www.w3.org/2000/svg" width="28.85" height="32" viewBox="0 0 256 284"><path fill="#F9AB00" d="M256.003 247.933a35.224 35.224 0 0 1-39.376 35.161c-18.044-2.67-31.266-18.371-30.826-36.606V36.845C185.365 18.591 198.62 2.881 216.687.24a35.221 35.221 0 0 1 39.316 35.16z"/><path fill="#E37400" d="M35.101 213.193c19.386 0 35.101 15.716 35.101 35.101c0 19.386-15.715 35.101-35.101 35.101S0 267.68 0 248.295c0-19.386 15.715-35.102 35.101-35.102m92.358-106.387c-19.477 1.068-34.59 17.406-34.137 36.908v94.285c0 25.588 11.259 41.122 27.755 44.433a35.161 35.161 0 0 0 42.146-34.56V142.089a35.222 35.222 0 0 0-35.764-35.282"/></svg>`,
-      scriptBundling: (options?: GoogleAnalyticsInput) => withQuery('https://www.googletagmanager.com/gtag/js', {
-        id: options?.id,
-      }),
-      import: {
-        name: 'useScriptGoogleAnalytics',
-        from: resolve('./runtime/registry/google-analytics'),
-      },
-    },
+    // {
```
huang-julien: @harlan-zw do you think we should remove this and its composable ?

huang-julien:
I think this is good for a first implementation. I'll probably open another PR for internal DX to ease the developement with TPC.

src/tpc/utils.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  chunks.push(`export type Input = RegistryScriptInput${hasParams ? '<typeof OptionSchema>' : ''}`)
+
+  chunks.push(`
+export function ${input.scriptFunctionName}<T extends ${input.tpcTypeImport}>(_options?: Input) {
+${functionBody.join('\n')}
+  return useRegistryScript${hasParams ? '<T, typeof OptionSchema>' : ''}('${input.tpcKey}', options => ({
+        scriptInput: {
+            async: true,
```
harlan-zw: this is already the default

package.json
```diff
@@ -54,9 +54,14 @@
...
       "@unhead/schema",
       "#nuxt-scripts",
-      "#nuxt-scripts-validator"
+      "#nuxt-scripts-validator",
+      "third-party-capital",
+      "knitwork"
     ]
   },
+  "peerDependencies": {
+    "third-party-capital": "^1.0.30"
```
harlan-zw: what is the rationale for this?

src/module.ts
```diff
@@ -74,6 +76,12 @@ export interface ModuleHooks {
   'scripts:registry': (registry: RegistryScripts) => void | Promise<void>
 }
 
+declare module '@nuxt/schema' {
```
harlan-zw, line 79: this shouldn't be needed if you've ran `pnpm dev:prepare`

src/module.ts
```diff
@@ -120,14 +128,23 @@ export default defineNuxtModule<ModuleOptions>({
       resolve('./runtime/registry'),
     ])
 
+    addImports({
```
harlan-zw: i'd prefer this to be an explicit import as end users wouldn't be using it, there's already an alias you can import from `#nuxt-scripts-utils`

src/module.ts
harlan-zw: although it seems a bit flakey so maybe it's needed to avoid tscheck issues, we should fix this in nuxt-module-builder at some point

src/registry.ts
```diff
@@ -4,28 +4,16 @@ import type { IntercomInput } from './runtime/registry/intercom'
 import type { SegmentInput } from './runtime/registry/segment'
 import type { NpmInput } from './runtime/registry/npm'
 import type { PlausibleAnalyticsInput } from './runtime/registry/plausible-analytics'
-import type { GoogleAnalyticsInput } from './runtime/registry/google-analytics'
+// import type { GoogleAnalyticsInput } from './runtime/registry/google-analytics'
```
harlan-zw: delete

src/tpc/google-analytics.ts
```diff
@@ -0,0 +1,54 @@
...
+        // allow dataLayer to be accessed on the server
+        stub: ({ fn }) => {
+          return fn === 'dataLayer' ? [] : undefined
+        },
+        tpcKey: 'gtag',
+        tpcTypeImport: 'GoogleAnalyticsApi',
+        augmentWindowTypes: true,
+      })
+    },
+    filename: 'tpc/google-analytics.ts',
```
harlan-zw: can we namespace to nuxt scripts so it's clear that this module is creating these files

harlan-zw:
I'm still not convinced TPC in its current format will be worth maintaining with Nuxt Scripts in the long term but I'm happy to move this forward for the time being once comments are addressed. 

Either way this is a nice solution to avoiding the runtime of TPC, great work @huang-julien :clap:

package.json
huang-julien: I moved it to peer dep so module managers won't enforce a specific version of tpc. User will be free to update tpc on their own. 

I'll change it to >=1

src/module.ts
huang-julien: yes :+1:

package.json
harlan-zw: What's the issue with having it as a dependency though? It's not clear

huang-julien: Having it as a dependency prevent users from having for example tpc v1.2.30. If set in peerDeps tpc should be updatable by users without tpc's version being locked by @nuxt/scripts

harlan-zw: Seems like a reasonable constraint to avoid breaking behaviours. If TPC introduces new scripts or config we'd need to add code to support them anyway.

Users can always override the version if they really need

PR #86, opened 2024-06-09
chore: migrate to Nuxt ESLint

shinGangan:
### ❓ Type of change

### 📚 Description

Repositories in Nuxt oranization are migrating to use [Nuxt ESLint](https://eslint.nuxt.com/).
Therefore, this repository has also been migrated to use `@nuxt/eslint-config` instead of `@antfu/eslint-config`.

#### Note

`node/prefer-global/process` and `node/prefer-global/buffer` are rules provided by `eslint-plugin-n`, and were not provided by default in `@nuxt/eslint-config`.

As before, if necessary, you can configure all `eslint-plugin-n` rules configured in `@antfu/eslint-config` in a separate PR. 👍 

I thought that this rules should be implemented after receiving @harlan-zw 's opinion, so it has not been addressed in this PR.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,9 +1,0 @@
-import antfu from '@antfu/eslint-config'
-
-export default antfu({
-  rules: {
-    'no-use-before-define': 'off',
-    'node/prefer-global/buffer': 'off',
-    'node/prefer-global/process': 'off',
-  },
-})
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -36,6 +36,7 @@ export default createConfigForNuxt({
       'vue/multi-word-component-names': 'off',
       'vue/no-v-html': 'off',
       'vue/require-default-prop': 'off',
+      'vue/no-multiple-template-root': 'off',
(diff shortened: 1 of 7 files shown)


shinGangan:
Hi @harlan-zw , please review and comments when you have a time 🙏

shinGangan:
Btw, I would like to take advantage of the monopo configuration to simplify package management in the `playground/` and `docs/`. 
Is this okay to try?
(I will implement this in another PR)

eslint.config.js
```diff
@@ -1,9 +1,43 @@
-import antfu from '@antfu/eslint-config'
+// @ts-check
+import { createConfigForNuxt } from '@nuxt/eslint-config/flat'
 
-export default antfu({
-  rules: {
-    'no-use-before-define': 'off',
```
shinGangan, line 5: ### Note

This rule is an additional rule specific to `@antfu/eslint-config`.

https://github.com/antfu/eslint-config/blob/main/src/configs/javascript.ts#L154

harlan-zw:
Thanks! This is some great work

PR #88, opened 2024-06-10
docs: add more detail to billing docs

kara:
Tiny PR to explain the trade-off if you use the static maps API (more up front if users engage, less if they don't). @harlan-zw This was my understanding from your comment - let me know if it's not right.


harlan-zw:
This is correct per the current logic. It may change when the caching is implemented but can tweak docs from there.

PR #99, opened 2024-06-16
refactor(tpc): move tests to AST instead of code snapshots

huang-julien:
This PR Refactor TPC tests. Snapshots slows way too much the develomement of TPC. It uses now AST to test the generation of TPC composables


harlan-zw:
I admire the code and effort gone into this, but personally, I think a human-readable snapshot is more useful for something that shouldn't be changing that frequently. It also looks like substantially more effort to maintain.

Given you've been maintaining this though, I am happy to merge, can you just fix the conflict.

huang-julien:
Thanks ! I'll regen the lockfile then merge it.

The reason behind removing snapshots is that any modification will force to update the snapshot. Moving it to AST allows more flexibility in the code generation without having to update for change.
I admit this is a bit more complex to read but we can remove it later if it is too complex for future contributors.

PR #108, opened 2024-06-23
refactor: refactor `import { type foo }` to `import type { foo }`

shinGangan:
### ❓ Type of change

### 📚 Description

`import type { foo }` and `import { type foo }` were mixed, so I refactored it to `import type { foo }`.

  - related: https://nuxt.com/blog/v3-8#type-import-changes

#### Note

Refactoring after the official release is hard, so I fixed it first.

### DIFF
--- a/docs/components/content/GoogleMapsDemo.vue
+++ b/docs/components/content/GoogleMapsDemo.vue
@@ -32,23 +32,11 @@ function handleReady(_map: Ref<google.maps.Map>) {
       />
     </div>
     <div class="text-center">
-      <UAlert
-        v-if="!isLoaded"
-        class="mb-5"
-        size="sm"
-        color="blue"
-        variant="soft"
-        title="Hover to load"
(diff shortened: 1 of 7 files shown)


harlan-zw:
Thanks! Ideally, the nuxt eslint rules would cover this.

shinGangan:
> Thanks! Ideally, the nuxt eslint rules would cover this.

I feel the same way.
Nuxt ESLint has rules set, so the problem may be with the typescript-eslint rules themselves.
(Can it not be detected in a case like this?🤔)

- [Nuxt ESLint: rules](https://github.com/nuxt/eslint/blob/417063ee6c2dd9f3c3c095d5e54b728024208e31/packages/eslint-config/src/flat/configs/typescript.ts#L64-L67)
- setting rules: https://typescript-eslint.io/rules/consistent-type-imports/

I'll also look into the rules of typescript-eslint when I have time.

PR #113, opened 2024-06-26
fix(tpc): custom feature detection flags

huang-julien:
### ❓ Type of change

### 📚 Description
This PR add feature detection for tpc composables. It also removes `augmentWindowTypes` to make it default


huang-julien (commented):
@flashdesignory do you need it in other components ?

src/runtime/composables/useFeatureDetection.ts
```diff
@@ -0,0 +1,13 @@
+import { onMounted } from 'vue'
+
+export function useFeatureDetection(id: string) {
+  onMounted(() => {
+    onMounted(() => {
```
flashdesignory: I think that might be a typo to have two onMounted() calls nested?

flashdesignory:
> @flashdesignory do you need it in other components ?

once GoogleMaps and Youtube is integrated, we should also add the feature detection there 🙏

flashdesignory (approved):
LGTM 🚀

src/runtime/composables/useFeatureDetection.ts
huang-julien: :joy: oops

huang-julien:
yes we need to discuss about theses components at the next meeting :)

harlan-zw:
Aren't these already implemented as part of https://github.com/nuxt/scripts/blob/main/src/runtime/composables/useScript.ts#L26 ?

flashdesignory:
we want to differentiate between all nuxt scripts and the third-party-capital integrations. 
Additionally, the naming convention that is currently used, diverged from the flags introduced in the initial pr. 
This results in a mismatch between nuxt script and chrome flags.

harlan-zw:
> we want to differentiate between all nuxt scripts and the third-party-capital integrations.

You can just use the key?

> Additionally, the naming convention that is currently used, diverged from the flags introduced in the initial pr.
> This results in a mismatch between nuxt script and chrome flags.

Yes, this was intentional. What / which chrome flags are you referring to?

flashdesignory:
We implemented flags in Chrome that we can use to pull usage data. 
These flags currently exist:

```
nuxt-third-parties-gtm
nuxt-third-parties-ga
nuxt-third-parties-GoogleMaps
nuxt-third-parties-YoutubeEmbed
```

harlan-zw:
> We implemented flags in Chrome that we can use to pull usage data. These flags currently exist:
> 
> ```
> nuxt-third-parties-gtm
> nuxt-third-parties-ga
> nuxt-third-parties-GoogleMaps
> nuxt-third-parties-YoutubeEmbed
> ```
> 
> previous pr: [nuxt/third-party-capital#53](https://github.com/nuxt/third-party-capital/pull/53)

Is it possible to update the flags? It would make more sense for the scope prefix to match the module name.

It's a bit clunky to try and support both naming conventions

huang-julien:
any progress on this ?

flashdesignory:
@huang-julien - thanks for following up.

I'd still like to use the updated naming convention for the following reasons:

1. we are more interested in the third-party-capital integrations 
2. updating flags that are already deployed in chrome is time intense
3. more granular usage reports available, if we add all Nuxt script flags to chrome in the future as well

Personally, I think the work is doable, since we're not dealing with a large number of third-parties anytime soon.
Keeping both group of flags updated seems reasonable.

huang-julien:
> the updated naming convention

Are you talking about `nuxt-third-parties-${id}` ? @harlan-zw wdyt if we move to this convention in `useScript` ?

flashdesignory:
> > the updated naming convention
> 
> Are you talking about `nuxt-third-parties-${id}` ? @harlan-zw wdyt if we move to this convention in `useScript` ?

correct, the only thing I am not sure about is if the ids match. 
We previously used "ga" and "gtm"

huang-julien:
Yes google analytics was using `gtag` instead of `ga`. It's now fixed but shouldn't the main script in the data in tpc have the same id used for feature detection ?

@harlan-zw i'm changing the feature detection to `nuxt-third-parties-${id}`, are you okay with it ?

harlan-zw:
> we are more interested in the third-party-capital integrations

This is fine, I don't see how using different keys stops you from doing this.

> updating flags that are already deployed in chrome is time intense

It seems like this is just a matter of changing a string but maybe you can clarify further exactly how much work are we talking?

> more granular usage reports available, if we add all Nuxt script flags to chrome in the future as well

Not sure how using different keys helps with this.

The reason why I prefer a properly named prefix:
- Chrome may not be the only ones using these flags
- Clear where the scripts are being loaded from

Keep in mind that even if we merge this, Youtube and Google Maps will still be broken.

flashdesignory:
Maybe I am missing something 🤷. 
If I understand the current implementations correctly,  a user has two options to use Google Analytics and Google Tag Manager: 

1. the default nuxt script version
2. the tpc integration (if the user opts into using that). 

Is that assumption correct?

If so, and I am using the same flags from nuxt script, how can I differentiate if a user uses 1) or 2) ?
Using the same flag wouldn't give me a correct signal. 
Adding an additional flag (with this pr), would allow me to just get the usage of the tpc integrations.

harlan-zw:
We only give the user one option of which to use, it would be confusing otherwise and add no value to the end user. 

GA and GTM is using TPC.

housseindjirdeh:
> It seems like this is just a matter of changing a string but maybe you can clarify further exactly how much work are we talking? Ideally these wouldn't of been set initially until we were close to making the module public, breaking changes should have been expected.

You're right, it's a [quick string change](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/timing/performance_mark.cc;l=157-161?q=nuxt-third-parties&ss=chromium). The larger concern is that any changes in Chromium will take a while before it hits stable (likely well after this module goes public and gathers adoption). We wanted to get a bit ahead of the curve by naming these strings before the module ships since it can take months to see the change in stable browser.

harlan-zw:
Thanks for clarifying, In this case, I'd consider this PR okay to go ahead with. Do we need to handle GTM as well @huang-julien

huang-julien:
tag manager has already `gtm` as key in tpc.

src/runtime/composables/useScript.ts
```diff
@@ -25,7 +25,7 @@ export function useScript<T extends Record<string | symbol, any>>(input: UseScri
     if (!nuxtApp._scripts?.[id]) {
       performance?.mark?.('mark_feature_usage', {
         detail: {
-          feature: `nuxt-scripts:${id}`,
+          feature: `nuxt-third-parties-${id}`,
```
harlan-zw: I'd prefer if the non-TPC scripts used the old key name.

So could we add a config key on NuxtUseScriptOption `performanceMarkFeature` which can be added as deprecated, then we provide the full key via the TPC scripts

```ts
feature: options?.performanceMarkFeature || `nuxt-scripts:${id}`,
```

src/tpc/utils.ts
```diff
@@ -86,7 +87,7 @@ declare global {
   chunks.push(`
 export function ${input.scriptFunctionName}<T extends ${input.tpcTypeImport}>(_options?: Input) {
 ${functionBody.join('\n')}
-  return useRegistryScript${hasParams ? '<T, typeof OptionSchema>' : ''}('${input.tpcKey}', options => ({
+  return useRegistryScript${hasParams ? '<T, typeof OptionSchema>' : ''}('${input.registryId ?? input.tpcKey}', options => ({
```
harlan-zw: we then pass in 

```ts
performanceMarkFeature: 'nuxt-third-parties-${id}`,
```

flashdesignory:
@huang-julien - if you're using the `key` from tpc as the `id`, do we need to handle the Google Analytics `key` ?

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/timing/performance_mark.cc;l=157-161?q=nuxt-third-parties&ss=chromium

https://github.com/GoogleChromeLabs/third-party-capital/blob/main/src/third-parties/google-analytics/data.json

I remember you mentioned that these are different and if that's the case, I think we can rename that one in tpc (@housseindjirdeh ).

src/runtime/composables/useScript.ts
huang-julien: i'm good with this  :+1:

huang-julien: as deprecated ? Maybe as internal if you don't want it to be public

harlan-zw (approved):

PR #124, opened 2024-07-03
fix: use object syntax to define NuxtConfigScriptRegistry

### LINKED ISSUE #92: Module Options accessed via runtime are broken
The latest [module-builder](https://github.com/nuxt/module-builder) does not allow us to import types from the runtime folder from the module.ts file.

This is breaking integrations using any values that are from the runtime folder such as `registry`.

huang-julien:
### 🔗 Linked issue

fix #92

### ❓ Type of change

### 📚 Description
This PR moves NuxtConfigScriptRegistry to Object syntax to correctly loop over ScriptRegistry keys, avoiding it to merge values types when using `Record`

### DIFF
--- a/src/runtime/types.ts
+++ b/src/runtime/types.ts
@@ -106,9 +106,9 @@ export interface ScriptRegistry {
 }
 
 export type NuxtConfigScriptRegistryEntry<T> = true | 'mock' | T | [T, NuxtUseScriptOptionsSerializable]
-export type NuxtConfigScriptRegistry<T extends keyof ScriptRegistry = keyof ScriptRegistry> = Partial<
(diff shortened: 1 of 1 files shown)


harlan-zw:
I'm not sure if this solves the issue. The problem can be seen here: https://unpkg.com/browse/@nuxt/scripts@0.5.1/dist/module.d.ts (second line)

All of the type imports are being transformed to .js imports from the latest nuxt-module-builder which broke cross runtime types. We need to move the types to build time and import into runtime I believe

harlan-zw (approved):

harlan-zw:
Good catch, looks to fix it :+1:

Issue #136, opened 2024-07-07
Unterminated regular expression error in 0.6.2


benfeather:
After upgrading @nuxt/scripts to 0.6.2 and updating my global scripts from an array to an object, I'm getting this error:

![CleanShot 2024-07-08 at 10 03 12@2x](https://github.com/nuxt/scripts/assets/43571410/e60d176c-ea4a-41a5-a05b-cd0e05a0100a)

Not sure Why the error lists the "YouTube embed" component, but the the error only occurs if @nuxt/scripts is enabled.

I have tried removing the "YouTube embed" component and I'm still getting the error, so I believe it's an issue with the scripts package.

benfeather:
That component is using the "useScriptTag" composable, not sure if that would cause the error, but it did the same thing when disabled, so not sure.

harlan-zw:
@huang-julien seems like a bug with the AST parsing of vue code which includes `useScript`. 

I think maybe we should catch the exception thrown from the transforms and allow any issues to propagate through the rest of the rollup plugins.

harlan-zw:
Thanks for reporting, should be fixed in 0.6.4.

PR #149, opened 2024-07-14
fix: safer top level await checks

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt/scripts/issues/136

### ❓ Type of change

### 📚 Description

The current implementation of top-level checks has some performance issues that I'd like to address while solving this bug.

Essentially we don't want to do any double-transforms of code, which is the cause of the issue and my performance concerns. This means I had to rip out most of the code as the new snippet we need to detect is equivalent to

```ts
const { $script } = useScript("/test.js");
[__temp, __restore] = _withAsyncContext(() => $script), await __temp, __restore();
```

This logic is not perfect and could do with a number of improvements, however if many people are running into this it can be solved at runtime quite easily (if awaiting in non-mounted context we can resolve the promise early).


huang-julien (commented):
This doesn't seems to check top-level await $script.load() anymore 🤔

PR #163, opened 2024-07-18
feat(GA,GTM): configurable dataLayer name

huang-julien:
### ❓ Type of change

### 📚 Description

Update to tpc 1.1.1 with minor changes. 

- adds a new dataLayerName option
- on our side we handle theses options and we define defaults for `dataLayerName` to avoid conflicts between gtm and ga
  - `defaultGa` for google analytics
  - `defaultGtm` for google tag manager


docs/content/scripts/analytics/google-analytics.md
```diff
@@ -27,7 +27,8 @@ export default defineNuxtConfig({
   scripts: {
     registry: {
       googleAnalytics: {
-        id: 'YOUR_ID'
+        id: 'YOUR_ID',
+        dataLayerName: 'defaultGa'
```
harlan-zw: end user should not need to define this, it should only be listed under docs for adding multiple google analytics

docs/content/scripts/analytics/google-analytics.md
```diff
@@ -67,6 +68,7 @@ export default defineNuxtConfig({
       scripts: {
         googleAnalytics: {
           id: '', // NUXT_PUBLIC_SCRIPTS_GOOGLE_ANALYTICS_ID
+          dataLayerName: 'defaultGa' // NUXT_PUBLIC_SCRIPTS_GOOGLE_ANALYTICS_DATALAYERNAME
```
harlan-zw: i think this gets title cased `NUXT_PUBLIC_SCRIPTS_GOOGLE_ANALYTICS_DATA_LAYER_NAME` (but I don't think we need it here)

docs/content/scripts/analytics/google-analytics.md
```diff
@@ -127,6 +130,10 @@ export const GoogleAnalyticsOptions = object({
    * The Google Analytics ID.
    */
   id: string(),
+  /**
+   * The datalayer's name you want it to be associated with
+   */
+  dataLayerName: string()
```
harlan-zw: optional

src/runtime/registry/google-tag-manager.ts
```diff
@@ -3,34 +3,36 @@ import { withQuery } from 'ufo'
 import type { GoogleTagManagerApi } from 'third-party-capital'
 import { useRegistryScript } from '#nuxt-scripts-utils'
 import type { RegistryScriptInput } from '#nuxt-scripts'
-import { object, string } from '#nuxt-scripts-validator'
+import { defu } from 'D:/repo/scripts/node_modules/.pnpm/defu@6.1.4/node_modules/defu/dist/defu.mjs'
```
flashdesignory: nit: this import looks weird, is that correct?

flashdesignory (approved):
LGTM, assuming the imports are good as is 👍

src/runtime/registry/google-tag-manager.ts
huang-julien: woooops 🤦‍♂️

scripts/generateTpcScripts.ts
```diff
@@ -13,32 +12,27 @@ export interface TpcDescriptor {
   tpcTypeImport: string
   key: string
   registry?: any
-  options: ({
```
harlan-zw, line 16: any particular reason to change this? this matches the `useRegistryScript` interface so keeps code closer aligned

src/runtime/registry/google-tag-manager.ts
```diff
@@ -1,36 +1,32 @@
...
 declare global {
   interface Window extends GoogleTagManagerApi {}
 }
 export type GoogleTagManagerInput = RegistryScriptInput<typeof GoogleTagManagerOptions>
 
 export function useScriptGoogleTagManager<T extends GoogleTagManagerApi>(_options?: GoogleTagManagerInput) {
+  _options = defu(_options, { dataLayerName: 'defaultGtm' })
   return useRegistryScript<T, typeof GoogleTagManagerOptions>(_options?.key || 'google-tag-manager', options => ({
     scriptInput: {
       src: withQuery('https://www.googletagmanager.com/gtm.js', { id: options?.id }),
```
harlan-zw, line 20: the src needs to include the dataLayer name otherwise how will it link it :thinking:

harlan-zw:
Personally, I'd prefer to see the default behaviour to use the default `window.dataLayer`, I think it's more intuitive for end users if they really need to pull for the underlying implementation. 

Is using `dataLayers` an existing best practice otherwise? A non-nested key would also simplify things, just using directly what the user provides. Again this is more intuitive for end users, we want to avoid hiding implementation details.

src/runtime/registry/google-tag-manager.ts
huang-julien: hmmm  it's not in TPC :eyes: . Is there a default name for data layers ? 
cc @flashdesignory

flashdesignory:
> Personally, I'd prefer to see the default behaviour to use `window.dataLayer`, I think it's more intuitive for end users if they really need to pull for the underlying implementation.
> 
> Is using `dataLayers` an existing best practice otherwise? A non-nested key would also simplify things, just using directly what the user provides. Again this is more intuitive for end users, we want to avoid hiding implementation details.

The idea of the `window.dataLayers` object was to keep it scoped, but it is less popular, compared to just assigning custom names directly to the windows object.

we can revert back to the previous implementation, where we'd use `window.dataLayer`, unless we get another dataLayer name passed in. TPC, will expect a value from the wrapper (even if it's the default name).

harlan-zw:
Oh my mistake, I didn't realize this is how it was baked into TPC. We can go with this solution, some doc would be good though. Will merge for now to move it along.

src/runtime/registry/google-tag-manager.ts
harlan-zw: Oh wait I just merged this and I forgot I left this comment.

Can we do some testing on it and make sure it's actually working? I'm pretty sure it will just use `dataLayer` if not configured.

cc: @flashdesignory

PR #176, opened 2024-07-22
fix(tpc): support customizing dataLayer variable

huang-julien:
### ❓ Type of change

> It's such a joy to work with Types🤷‍♂️ Julien Huang - 2024

With the latest and futur changes of TPC, it is better to make the API type of tpc composables depending on the return type of the use function instead of the API provided by TPC since it will differ for implementations reasons. This PR will also fixes a type issue with datalayer since it does not exist anymore in type provided by TPC (datalayers exist tho).


src/runtime/registry/google-tag-manager.ts
```diff
@@ -13,15 +13,19 @@ declare global {
 }
 export type GoogleTagManagerInput = RegistryScriptInput<typeof GoogleTagManagerOptions>
 
-export function useScriptGoogleTagManager<T extends GoogleTagManagerApi>(_options?: GoogleTagManagerInput) {
+function use(options: GoogleTagManagerInput) {
+  return { dataLayer: window.dataLayers[options.dataLayerName!], google_tag_manager: window.google_tag_manager }
+}
+
+export function useScriptGoogleTagManager(_options?: GoogleTagManagerInput) {
```
harlan-zw, line 21: do we need to bump TPC to fix these type errors or something?

huang-julien: yes, i'll put this in draft mode

scripts/generateTpcScripts.ts
```diff
@@ -28,13 +29,14 @@ const scripts: Array<TpcDescriptor> = [
...
+    tpcTypeAugmentation: 'GoogleTagManagerApi',
+    tpcTypesImport: ['DataLayer'],
     key: 'googleTagManager',
     performanceMarkFeature: 'nuxt-third-parties-gtm',
-    returnUse: '{ dataLayer: window.dataLayers[options.dataLayerName!], google_tag_manager: window.google_tag_manager }',
+    returnUse: '{ dataLayer: window[options.l!] as DataLayer, google_tag_manager: window.google_tag_manager }',
     returnStub: 'fn === \'dataLayer\' ? [] : void 0',
     defaultOptions: {
-      dataLayerName: 'defaultGtm',
+      l: 'defaultGtm',
```
harlan-zw: this should be undefined by default so we aren't adding it to the URL unless it's specified, we fallback to `dataLayer` for both GA and GTM

scripts/utils.ts
```diff
@@ -24,12 +24,39 @@ export async function generateTpcContent(input: TpcDescriptor) {
...
+    tpcTypes.add(input.tpcTypeAugmentation)
+
+    chunks.push(`
+    declare global {
+      interface Window extends ${input.tpcTypeAugmentation} {}
+    }`)
+  }
+  else {
+    chunks.push(`
+    declare global {
```
harlan-zw: i don't think this is a good idea

if we can't make the window / internal types work for `l` then just ts ignore them, as long as the composable has the correct types

src/runtime/registry/google-analytics.ts
```diff
@@ -1,32 +1,38 @@
...
 }
+export const GoogleAnalyticsOptions = object({ id: string(), l: optional(string()) })
 export type GoogleAnalyticsInput = RegistryScriptInput<typeof GoogleAnalyticsOptions>
 
-export function useScriptGoogleAnalytics<T extends GoogleAnalyticsApi>(_options?: GoogleAnalyticsInput) {
-  _options = defu(_options, { dataLayerName: 'defaultGa' })
-  return useRegistryScript<T, typeof GoogleAnalyticsOptions>(_options?.key || 'googleAnalytics', options => ({
+function use(options: GoogleAnalyticsInput) {
+  return { dataLayer: window[options.l!] as DataLayer,
+    gtag(...args: any) { (window[options.l!] as DataLayer).push(args) } }
```
harlan-zw: we had this typed before

housseindjirdeh (approved):

src/runtime/registry/google-analytics.ts
huang-julien: Are you talking about gtag() ? Not anymore with tpc 2.0. So it's better to rely on our own use function

huang-julien:
waiting for https://github.com/GoogleChromeLabs/third-party-capital/pull/59 (no breaking changes) to remove default options (`l`)

src/runtime/registry/google-analytics.ts
harlan-zw: ye gtag sorry, I agree it's better but can we re-add the types

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/gtag.js/index.d.ts

huang-julien: `DataLayer` type from tpc conflict with `@types/gtag.js` GtagCommand types. 

cc @flashdesignory

flashdesignory: @huang-julien - I'm happy to adjust the DataLayer type as needed. Do you want to open a pr or what's the easiest / fastest to get you what you need?

flashdesignory: @huang-julien - not sure exactly where the conflict is happening, but could you use the GTag type from tpc?

maybe something like:
```
import type { DataLayer, GTag } from 'third-party-capital'

function use(options: GoogleAnalyticsInput) {
  const gtag: GTag = function (...args: any) {
    ((window as any)[options.l ?? 'dataLayer'] as DataLayer)?.push(args)
  }
  return { dataLayer: (window as any)[options.l!] as DataLayer,
    gtag,
  }
}
```

flashdesignory (approved):
LGTM

harlan-zw (approved):

PR #216, opened 2024-08-22
docs: use correct trigger name

manniL:
### ❓ Type of change


harlan-zw:
`mouseover` is correct I beleive? only the intersection observer and idle triggers have these special aliases, everything else is just the js dom event https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseover_event

manniL:
Ah, my bad! You are right 😋

PR #235, opened 2024-09-02
fix: update youtube placeholder default image url

### LINKED ISSUE #234: YouTube placeholder cannot show some videos
### 🐛 The bug

The current youtube placeholder url `https://i.ytimg.com/vi_webp/${props.videoId}/sddefault.webp` cannot show some videos.

#### `vi_webp` is not guaranteed to exist
https://i.ytimg.com/vi_webp/mJ8tq8AnNis/default.webp ([video](https://www.youtube.com/watch?v=mJ8tq8AnNis))

#### `sddefault` is not guaranteed to exist
https://i.ytimg.com/vi/jNQXAC9IVRw/sddefault.jpg ([video](https://www.youtube.com/watch?v=jNQXAC9IVRw))

### 🛠️ To reproduce

https://stackblitz.com/edit/nuxt-starter-lnbkmh?file=pages%2Findex.vue

### 🌈 Expected behavior

Expected the youtube placeholder can show all videos preview thumbnail image.

### ℹ️ Additional context

_No response_

serkodev:
### 🔗 Linked issue

Fix #234

### ❓ Type of change

### 📚 Description

Update the youtube placeholder url to `https://i.ytimg.com/vi/${props.videoId}/hqdefault.jpg`.
This thumbnail image url is guaranteed to exist. ([Reference](https://stackoverflow.com/a/20542029/10603887))


harlan-zw:
I think this is a reasonable solution however we need to be mindful as this may break the look and feel of current users videos given the size difference, the current is 640x480 with this change it becomes 480x360. 

Do you have any ideas on how to workaround this? I think the ideal case is probably two things:
1. props to easily choose the right thumbnail size / webp
2. a `<img onerror` fallback that serves the lower res safe size (if it's using the defaults). It would be nice to use picture or something for this but seems like it wouldn't work (https://github.com/whatwg/html/issues/8916)

manniL:
Also I'd still go with the `webp` endpoint (smaller image size)

serkodev:
If we provide props to choose the thumbnail size or WebP format, I believe we don't have a perfect solution to detect if a YouTube video doesn't support these options.

1. Use `fetch` to request the image and check the response code (we all know this isn't a good approach).
2. Fetch the YouTube Data API to get a list of thumbnail URLs (this method requires an API key and is not very client-side rendering-friendly).

There's a somewhat hacky method that involves detecting the image size.  
Since all 404 images will have a size of 120x90, we can use this characteristic to detect whether a thumbnail exists. If we detect that the image size is 120x90, we can then fallback to `default.jpg`.

The only drawback is that if the original image request (e.g., `default.jpg`, `1.jpg`, `2.jpg`, `3.jpg`) already has a size of 120x90, we can't determine if it's actually a 404. It may be forcibly fallback to `default.jpg`, unless we calculate the image data (e.g., hash) to determine if it's a not-found image, but that would make things increasingly complicated.

However, if we only provide props for thumbnail size or WebP format and let users determine whether the image is supported, then we might not need to detect if it's a 404 at all.

That's all I can think of for now. 🤣

harlan-zw:
I think the dimension heuristic is good enough

```html
<img src="https://i.ytimg.com/vi_webp/jNQXAC9IVRw/sddefault.webp" onload="if (this.naturalWidth == 120 && this.naturalHeight == 90) { (this.src = `https://i.ytimg.com/vi/jNQXAC9IVRw/hqdefault.jpg`) }" />
```

We should warn the developer in development that they should use the props when this fallback occurs.

serkodev:
Thanks for the feedback. So now Thank you for your feedback. Now, we also need to discuss the behavior of the `placeholder` slot prop. Should we keep passing in `sddefault.webp`?  
Do we need to add a `placeholderFallback` (to pass hqdefault.jpg) slot prop?

src/runtime/components/ScriptYouTubePlayer.vue
```diff
@@ -120,7 +120,9 @@ const rootAttrs = computed(() => {
   }) as HTMLAttributes
 })
 
+const fallbackPlaceHolder = computed(() => `https://i.ytimg.com/vi/${props.videoId}/hqdefault.jpg`)
 const placeholder = computed(() => `https://i.ytimg.com/vi_webp/${props.videoId}/sddefault.webp`)
+const isFallbackPlaceHolder = ref(false)
```
serkodev, line 126:
```suggestion
type YouTubeThumbnailType =
  // 120x90
  '1' | '2' | '3' | 'default' |
  // 320x180
  'mq1' | 'mq2' | 'mq3' | 'mqdefault' |
  // 480x360
  '0' | 'hq1' | 'hq2' | 'hq3' | 'hqdefault' |
  // 640x480
  'sd1' | 'sd2' | 'sd3' | 'sddefault' |
  // 1280x720
  'hq720' |
  // 1920x1080
  'maxresdefault'

function getYouTubeThumbnail(type: YouTubeThumbnailType, webp = false) {
  return `https://i.ytimg.com/${webp ? 'vi_webp' : 'vi'}/${props.videoId}/${type}.${webp ? 'webp' : 'jpg'}`
}
```

Just draft a function to get YouTube thumbnail here. I think we can use this for `placeholder` or provide for user to use?

harlan-zw:
Continuing here https://github.com/nuxt/scripts/pull/376

PR #248, opened 2024-09-05
fix(adsense)!: remove default `data-ad-format`

### LINKED ISSUE #247: Google Adsense: Fixed size display unit
### 🆒 Your use case

When I use `<ScriptGoogleAdsense/>` without the `data-ad-format` attribute, I expect it to render a [fixed size ad unit](https://developers.google.com/adsense/platforms/transparent/ad-tags#display_ad_unit_code_fixed_size).

But after some digging it seems that the `data-ad-format` has a default value of `auto` making the ad unit responsive. ([link](https://github.com/nuxt/scripts/blob/main/src/runtime/components/ScriptGoogleAdsense.vue#L7C1-L19C3))

### 🆕 The solution you'd like

I'd suggest either removing the ad format default or removing the defaults entirely.

_No response_

### ℹ️ Additional info

_No response_

asokawotulo:
### 🔗 Linked issue

Resolves #247

### ❓ Type of change

- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Removed the default ad format, allowing for fixed-size ad units

### DIFF
--- a/src/runtime/components/ScriptGoogleAdsense.vue
+++ b/src/runtime/components/ScriptGoogleAdsense.vue
@@ -14,7 +14,6 @@ const props = withDefaults(defineProps<{
    */
   trigger?: ElementScriptTrigger
 }>(), {
-  dataAdFormat: 'auto',
   dataFullWidthResponsive: true,
 })
 


src/runtime/components/ScriptGoogleAdsense.vue
```diff
@@ -37,9 +36,7 @@ const instance = useScriptGoogleAdsense({
 const { status } = instance
 
 onMounted(() => {
-  callOnce(() => {
```
harlan-zw: I did do this for a reason :thinking: 

Possibly because in dev the HMR was triggering a new ad each time. 

We might need to track which ads we've pushed an object for if we're going to remove this

src/runtime/components/ScriptGoogleAdsense.vue
```diff
@@ -14,7 +14,6 @@ const props = withDefaults(defineProps<{
    */
   trigger?: ElementScriptTrigger
 }>(), {
-  dataAdFormat: 'auto',
```
harlan-zw, line 17: I'm okay with this change but as it's a breaking change it would need to wait until 0.9.0 so maybe better to have it in a separate PR.

asokawotulo: Alright let me revise the PR

src/runtime/components/ScriptGoogleAdsense.vue
asokawotulo: I see, I'll try play around and see if I can come up with a solution for this

asokawotulo: @harlan-zw Just an idea, what if we use `callOnce` only in dev, but in production, we push every time the component is mounted.

Something along the lines of:
```
function pushAdSlot() {
  (window.adsbygoogle = window.adsbygoogle || []).push({})
}

onMounted(() => {
  import.meta.dev
    ? callOnce(() => {
        pushAdSlot()
      })
    : pushAdSlot()

  // ...
})
```

PR #277, opened 2024-09-25
fix(segment): proxySrc, fileName for proxying

georgekarapi:
### ❓ Type of change

### 📚 Description

Segment can be blocked through AdBlockers. One way to tackle it is using Self-hosted custom proxies (as mentioned [here](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/custom-proxy/)).

This PR introduces 2 options in Segment's config schema.

### DIFF
--- a/docs/content/scripts/tracking/segment.md
+++ b/docs/content/scripts/tracking/segment.md
@@ -115,6 +115,8 @@ You must provide the options when setting up the script for the first time.
 export const SegmentOptions = object({
   writeKey: string(),
   analyticsKey: optional(string()),
+  proxySrc: optional(string()),
+  fileName: optional(string()),
 })
 ```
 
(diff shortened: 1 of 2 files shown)


harlan-zw:
I don't think this quite makes sense as these aren't official Segment options, but maybe you can confirm with me.

As far as I know, there's already two workarounds available for this issue:

1. Use `bundle: true`, this will offload the script to be served by your Nuxt app, see (https://scripts.nuxt.com/docs/guides/bundling). This is by far the easiest option.
2. Set your own src using `useScriptSegment({ scriptInput: { src: 'https://cdn.segment-proxy.com/analytics.js' } })`

I'm not too sure where the  proxy API tracking calls stubbing comes in as this seems embedded in the script itself.

PR #287, opened 2024-10-04
fix: optional Valibot schema

huang-julien:
### ❓ Type of change

### 📚 Description

@harlan-zw following the discord discussion, this could be a first solution for TPC V3. This would allow to have typesafety but without runtime validation so we don't have to know how TPC or any other type is defined but passing it as a Type Parameter would be enough to warn devs that something is wrong

cc @flashdesignory


huang-julien:
The other one is to read the `d.ts` file exported from TPC, retrieve a TS AST then generate `valibot`'s validation schema with the AST

PR #290, opened 2024-10-08
feat: add `useScriptPaypal`

OrbisK:
> [!NOTE]  
> This PR is still WIP, opening for (hopefully) early feedback
### ❓ Type of change

### 📚 Description

This PR is to add a new script for paypal.

I have tried my hand at writing a script for paypal. There are a few points where I am still unsure. On the one hand with my current implementation of `useScriptPaypal()`, on the other hand with the possible splitting into different componetns. E.g. `<ScriptPaypalButtons>`. I have roughly listed a few things below that I would appreciate feedback on.

1. is it better to load the paypal script from the npm registry or via the url? (versioning?)
2. `useScriptPayPal` or `useScriptPaypal`
3. query serialization (keys with kebab-case, arrays)
4. maybe it makes sense to split the single paypal component into different possible components like `<ScriptPaypalButtons />` and/or <ScriptPaypalMessages />`. To get the source code for buttons, the query `components=buttons` is passed, for messages `components=messages`. However, it would probably make more sense to pass `components=buttons,messages` and only make the request once. How could this look structurally?

I would then add documentation as progress continues

> [!NOTE]  
> src/runtime/components/ScriptPayPal.vue is currently just for testing paypal buttons.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "@nuxt/eslint-config": "^0.5.7",
     "@nuxt/module-builder": "^0.8.4",
     "@nuxt/test-utils": "3.14.2",
+    "@paypal/paypal-js": "^8.1.2",
     "@types/semver": "^7.5.8",
     "@typescript-eslint/typescript-estree": "^8.7.0",
     "@unhead/schema": "1.11.6",
--- a/playground/pages/third-parties/paypal/nuxt-scripts.vue
+++ b/playground/pages/third-parties/paypal/nuxt-scripts.vue
@@ -1,0 +1,8 @@
+<template>
+  <ScriptPayPal
+    class="border border-gray-200 dark:border-gray-800 rounded-lg bg-white dark:bg-white/5"
+  />
+</template>
+
+<script setup lang="ts">
+</script>
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -90,7 +90,7 @@ importers:
         version: 3.7.0
       third-party-capital:
         specifier: 2.3.0
-        version: 3.0.0
+        version: 2.3.0
       ufo:
         specifier: ^1.5.4
         version: 1.5.4
@@ -116,6 +116,9 @@ importers:
       '@nuxt/test-utils':
         specifier: 3.14.2
         version: 3.14.2(h3@1.12.0)(magicast@0.3.5)(nitropack@2.9.7(magicast@0.3.5)(webpack-sources@3.2.3))(playwright-core@1.47.2)(rollup@4.21.3)(vite@5.4.5(@types/node@22.7.3)(terser@5.34.0))(vitest@2.1.1(@types/node@22.7.3)(terser@5.34.0))(vue-router@4.4.5(vue@3.5.9(typescript@5.6.2)))(vue@3.5.9(typescript@5.6.2))(webpack-sources@3.2.3)
+      '@paypal/paypal-js':
+        specifier: ^8.1.2
+        version: 8.1.2
       '@types/semver':
         specifier: ^7.5.8
         version: 7.5.8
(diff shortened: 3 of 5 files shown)


harlan-zw:
Hi, thank you for your hard work on this and sorry for the delay.

> 1. is it better to load the paypal script from the npm registry or via the url? (versioning?)

We want to try and avoid adding node_modules for end users for scripts they don't need as the module size will grow rapidly, so we prefer loading via URL. This gets a bit tricky for the types though.

> useScriptPayPal or useScriptPaypal

`useScriptPayPal` 

> 3. query serialization (keys with kebab-case, arrays)

Hmm there may be ways to clean this up but if this is working then I'm fine with what you have.

> 4. maybe it makes sense to split the single paypal component into different possible component

If we have or may have multiple components for one provider I think better to name them clearer. I'm not exactly sure if I follow what your saying otherwise as I haven't implemented the PayPal button myself, happy to defer to your judgement there.

OrbisK:
> If we have or may have multiple components for one provider I think better to name them clearer. I'm not exactly sure if I follow what your saying otherwise as I haven't implemented the PayPal button myself, happy to defer to your judgement there.

Unfortunately I have to ask you more specifically, not about the paypal script, but about the way paypal implements it. 

The same script may have several functionalities, depending on which one is requested. But this also means that if I want to use different paypal components in two places, then I would request 90% of the code twice. However, it would also be possible to simply pass both components directly to the request.

Another possibility would be to always request all “components” from paypal, but then you would have loaded 20% more script than necessary from paypal if you only use one.

It would be best if you had a way of knowing which components there should be before the specific request and thus customize the query parameters granularly.

Hope that makes sense so far.

OrbisK:
When I thought more about it, I decided to simply request all components first. Maybe you could think about opting out via config

OrbisK:
> Hi, thank you for your hard work on this and sorry for the delay. Overall is looking great.
> 
> > 1. is it better to load the paypal script from the npm registry or via the url? (versioning?)
> 
> We want to try and avoid adding node_modules for end users for scripts they don't need as the module size will grow rapidly, so we prefer loading via URL. This gets a bit tricky for the types though.
> 
> > useScriptPayPal or useScriptPaypal
> 
> `useScriptPayPal`
> 
> > 3. query serialization (keys with kebab-case, arrays)
> 
> Hmm there may be ways to clean this up but if this is working then I'm fine with what you have.
> 
> > 4. maybe it makes sense to split the single paypal component into different possible component
> 
> If we have or may have multiple components for one provider I think better to name them clearer. I'm not exactly sure if I follow what your saying otherwise as I haven't implemented the PayPal button myself, happy to defer to your judgement there.

okay, the paypal sdk is completely cursed. but i think i have a theoretical progress that shows in which direction it goes. I had not yet done the renaming to `usePayPal`. I would tackle that and the docs soon. 

Unfortunately I don't have a concrete project at the moment where I can test live. The paypal docs are also very wild 😃 some of the types don't match the docs either. 

I have worked around the problem with the components as described above.

if you say that everything still fits so far, I would then catch up on the refactoring and docs and try to use everything live in a project 😄

vincentdorian:
Is this still being worked on?

harlan-zw:
I can help out with this but it was so long ago I forgot what we were blocked on.

If you're happy with the state of it @OrbisK I'm happy to merge

harlan-zw:
- Very happy to ship incrementally
- Nightly is setup

> npm i https://pkg.pr.new/nuxt/scripts/@nuxt/scripts@e49434e

- I've given you write access @OrbisK if it helps speed anything along

harlan-zw:
cool let's close this one for now then

OrbisK:
@vincentdorian @Deeyuude If you want you can test this with: https://github.com/nuxt/scripts/pull/487 > https://pkg.pr.new/nuxt/scripts/@nuxt/scripts@487

PR #300, opened 2024-10-13
fix(stripe): prefer `@stripe/stripe-js` over `@types/stripe-v3`

### LINKED ISSUE #299: Stripe: Load TypeScript types from stripe-js, NOT from @types
### 🆒 Your use case

I just figured out that the types dont contain [stripe.initEmbeddedCheckout()](https://docs.stripe.com/js/embedded_checkout/init?type=expressCheckoutElement). When looking for the type definitions I saw that the repo for `@types/stripe-v3` doesnt exist anymore, last publish was 8 months ago.
 
The docs say that `@stripe/stripe-js` already contains types:
https://github.com/stripe/stripe-js?tab=readme-ov-file#typescript-support

It seems to be that they dropped support for `@types/stripe-v3`. Therefor, can we replace type definitions from the official package please?

huang-julien:
### 🔗 Linked issue

fix #299

### ❓ Type of change


harlan-zw:
Thanks for the PR, I'm thinking over the best plan for this as the module has so far managed to avoid any explicit third-party runtime node_module dependency code.

If we compare the package sizes we get:
- `@stripe/stripe-js` - 904KB
- `@types/stripe-v3` - 76KB

While this change is pretty negligible itself, if we extrapolate this to several other packages we may end up with adding ~20-30mb of node_module dependencies that aren't being used.

For now let's merge and for the v1 we can prompt the user to install these when used.

PR #366, opened 2025-01-30
feat(google-adsense): add Auto Ads support and improve script injection

nexoscreator:
### 🔗 Linked issue

No

### ❓ Type of change

#### Code Changes  
- Added `autoAds` option to enable or disable Auto Ads dynamically.  
- Modified `scriptInput.src` to append `?client=ca-pub-XXXXXX` dynamically.  
- Injected an inline script when `autoAds` is enabled to support automatic ad placement.  
- Improved TypeScript types for better validation and usability.  

#### Documentation Updates  
- Updated the **Google Adsense** documentation to include `autoAds` configuration.  
- Added a **Global Configuration** section in `nuxt.config.ts` to illustrate usage.  
- Expanded **Component API** details with  `data-ad-format` and `data-ad-layout` for in-article and in-feed ads.  
- Provided site verification instructions using a meta tag and `ads.txt` file.  

### 📷 Screenshots (if applicable)  
N/A  

### ✅ Checklist  

- [x] My code follows the project’s coding style and guidelines.  
- [x] I have updated the documentation accordingly.  
- [x] I have tested the changes locally to ensure they work as expected.  
- [x] I have checked for potential breaking changes.


harlan-zw:
Thank you for the work on this PR, as I don't use Google Adsense myself your feedback is really helpful.

All looks good although I'm a bit confused on "Auto Ads". Is this something from Google Ad vocabulary or is this something you've created?

It seems like we can possibly just always init the `adsbygoogle` array? Do you think this would cause any issues? Although end users should be able to configure `enable_page_level_ads` explicitly if this does something.

src/runtime/components/ScriptGoogleAdsense.vue
```diff
@@ -7,14 +7,16 @@ import type { ElementScriptTrigger } from '#nuxt-scripts/types'
...
+  dataAdFormat?: 'auto' | 'rectangle' | 'vertical' | 'horizontal' | 'fluid' | 'autorelaxed'
+  dataAdLayout?: 'in-article' | 'in-feed' | 'fixed'
   dataFullWidthResponsive?: boolean
   /**
    * Defines the trigger event to load the script.
    */
   trigger?: ElementScriptTrigger
 }>(), {
   dataFullWidthResponsive: true,
+  dataAdFormat: 'auto', // Default to Auto Ads
```
harlan-zw: Would this cause issues for users who didn't set it previously?

docs/content/scripts/ads/google-adsense.md
```diff
@@ -88,8 +124,16 @@ The generated meta tag will look like this:
 <meta name="google-adsense-account" content="ca-pub-<your-id>">
 ```
 
+Alternatively, add an `ads.txt` file to your `public` directory:
```
harlan-zw: I think you need to make it clearer that the end user needs to put their own input here possibly? Can you document where they'd find this info or what they need to replace

nexoscreator:
> Thank you for the work on this PR, as I don't use Google Adsense myself your feedback is really helpful.
> 
> All looks good although I'm a bit confused on "Auto Ads". Is this something from Google Ad vocabulary or is this something you've created?
> 
> It seems like we can possibly just always init the `adsbygoogle` array? Do you think this would cause any issues? Although end users should be able to configure `enable_page_level_ads` explicitly if this does something.


**Thank you for reviewing my PR!**  

Regarding your question about **"Auto Ads"**, it is an official term used by Google AdSense. Auto Ads automatically place ads on a page based on Google's algorithm. Here is the official documentation for reference: [Google AdSense Auto Ads](https://support.google.com/adsense/answer/9274025?hl=en).  

As for initializing the `adsbygoogle` array, I believe it should be safe to always initialize it. However, enabling `enable_page_level_ads` explicitly might be preferable to give end users more control over whether they want Auto Ads or not. If we initialize the array unconditionally, it shouldn't cause issues, but it would be good to clarify whether all users would expect this behavior.  

Let me know if you have any concerns, and I'm happy to adjust the PR accordingly! 😊🚀

src/runtime/components/ScriptGoogleAdsense.vue
nexoscreator: That's a great point! Setting `dataAdFormat` to `'auto'` by default might unintentionally override users who previously relied on the default behavior or manually set formats like `horizontal`, `vertical`, or `fluid`.  

To avoid potential issues, we could:  
- **Only set `'auto'` if `dataAdFormat` is undefined**, ensuring existing users aren’t affected.

nexoscreator: I've updated the code to **default** to `undefined` instead of `'auto'`, so that users who haven’t explicitly set `dataAdFormat` will retain the previous behavior.

docs/content/scripts/ads/google-adsense.md
nexoscreator: You're absolutely right—it's important to clarify where the user can find their Google AdSense Publisher ID and how they should replace the `<your-id>` placeholder. I've updated the documentation to include clear instructions on how users can find their Publisher ID in their AdSense account and how to use it in the meta tag or `ads.txt` file.

PR #378, opened 2025-02-08
fix: expect type error for optional type deps

ZTL-UwU:
### ❓ Type of change

### 📚 Description

Since the type dependencies were dropped in commit f54552654d047bf33eeef2a158819b78cdd60518, typechecking in projects that rely on Nuxt scripts is failing.

To get things moving again, I’ve added @ts-nocheck to ignore the type errors for now. I know this isn’t a great solution and can hide real issues, but it’s the best workaround I could come up with at the moment.


harlan-zw:
Thanks!

PR #390, opened 2025-02-13
docs: clarify async behavior and default script attributes

### LINKED ISSUE #304: Documented default script attributes vary from actual
### 🐛 The bug

The following script
```js
useScript('https://cdn.jsdelivr.net/npm/js-confetti@0.12.0/dist/js-confetti.browser.js')
```

currently yields
```html
<script data-onload="" data-onerror="" defer="" fetchpriority="low" crossorigin="anonymous" referrerpolicy="no-referrer" src="https://cdn.jsdelivr.net/npm/js-confetti@0.12.0/dist/js-confetti.browser.js" data-hid="5c23421"></script>
```

While [documentation](https://scripts.nuxt.com/docs/guides/key-concepts#default-behavior) lists `async` as one of the attributes that will be added by default, it is missing from the generated output above.

At the same time `fetchpriority="low"` seems to be missing from documentation, which is also a [default](https://github.com/unjs/unhead/blob/cd3c4caa7ca032de74b30ac90c38eb1638802908/packages/unhead/src/composables/useScript.ts#L126).

### 🛠️ To reproduce

N/A

### 🌈 Expected behavior

🌈

### ℹ️ Additional context

_No response_

nexoscreator:
### 🔗 Linked issue

Resolves #304  : Some users expected `async` to be applied by default, leading to confusion. This update makes the behavior explicit.

### ❓ Type of change

### 📚 Description

- Updated documentation to clarify that `async` is not applied by default   because `defer` is used.  
- Added a note explaining how to explicitly enable `async` by disabling `defer`.  
- Improved readability of default script attributes section.

### DIFF
--- a/docs/content/docs/1.guides/0.key-concepts.md
+++ b/docs/content/docs/1.guides/0.key-concepts.md
@@ -39,12 +39,14 @@ to the hydration process. Instead, scripts are loaded by default when Nuxt is fu
 
 You can change this behavior by modifying the [defaultScriptOptions](/docs/api/nuxt-config#defaultscriptoptions).
 
-Nuxt Scripts will also insert several extra tags to the script element to help with performance and privacy.
+Nuxt Scripts will also insert several extra tags to the `<script>` element to optimize performance and privacy.
 - `async` - Scripts are loaded asynchronously to prevent blocking the rendering of the page.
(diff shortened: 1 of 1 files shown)


harlan-zw:
Thanks, this is an improvement but there's still some issues, like fetchpriority is conditional added based on the trigger but let's go with this for now.

PR #392, opened 2025-02-16
feat(gtm): add onBeforeGtmStart

### LINKED ISSUE #243: Add consent mode for Google Analytics registry script
### 🆒 Your use case

The current Google Analytics registry does not use [consent mode](https://developers.google.com/tag-platform/security/guides/consent?consentmode=advanced). This is used for communicating the user's consent status to Google.

### 🆕 The solution you'd like

I would like to have it as part of the options for the registry script so I can easily define default consent settings as well as the option of opting into consent mode v2.

Currently I have opted to just copy the registry script and redefine it with an expanded clientInit property since the default consent needs to be set before `gtag('js', ...)` and `gtag('config', ...)` is called.

```ts
[...]
clientInit: import.meta.server
    ? undefined
    : () => {
        window.dataLayer = window.dataLayer || []
        window.gtag = function () {
          // eslint-disable-next-line prefer-rest-params
          window.dataLayer.push(arguments)
        }

        window.gtag('consent', 'default', {
          ad_user_data: 'denied',
          ad_personalization: 'denied',
          ad_storage: 'denied',
          analytics_storage: 'denied',
        })
        window.gtag('set', { ads_data_redaction: 'true' })

        window.gtag('js', new Date())
        window.gtag('config', website.analytics4Key)
      },
[...]
```

### ℹ️ Additional info

_No response_

huang-julien:
### 🔗 Linked issue
resolve #243

### ❓ Type of change

### 📚 Description
This PR improves GTM by adding a new callback to allow users to push anything into the datalayer before we push gtm.start . This way, users will be able to correctly set the consent mode

### DIFF
--- a/docs/content/scripts/tracking/google-tag-manager.md
+++ b/docs/content/scripts/tracking/google-tag-manager.md
@@ -128,10 +128,16 @@ export const GoogleTagManagerOptions = object({
   id: string(),
   /**
    * The name of the dataLayer you want to use
-   * @default 'defaultGtm'
+   * @default 'dataLayer'
    */
   dataLayerName: optional(string())
 })
+```
+
+### Options types
+
+```ts
+type GoogleTagManagerInput = typeof GoogleTagManagerOptions & { onBeforeGtmStart?: ((gtag: Gtag) => void) => void }
 ```
 
 ## Example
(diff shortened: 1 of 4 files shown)


harlan-zw:
This is a great escape hatch. Would it be possible to document with an example of when you'd use this :pray:

harlan-zw (approved):

huang-julien:
Seems like we DO need to use a `gtag()` fn in order to configure the consent mode.... This makes 0 sense.

huang-julien:
> I think if all EU users need the consent mode we should have a dedicated docs section for that 🤔

Are you talking about a whole page ?

harlan-zw:
Nah just like a section in the GTM page, let's go with this for now. Thanks!

Issue #404, opened 2025-02-25
Allow changing the Google Tag Manager script src


Reyno-Skeps:
### 🆒 Your use case

For server-side tagging a custom src needs to be set for Google Tag Manager.

Currently this src is hard coded in: https://github.com/nuxt/scripts/blob/f96f7b51549f302160559041bbf056589e899b04/src/runtime/registry/google-tag-manager.ts#L47

Whenever we want to change the src manually the ID is no longer appended, make it so we have to define the ID twice.

```
googleTagManager: {
	id: 'GTM-123456',
	scriptInput: {
		src: 'https://gtm.example.com/gtm.js?id=GTM-123456',
	},
},
```

### 🆕 The solution you'd like

A simple solution to the problem could be to add an option for the src.

`src: withQuery(options?.src || 'https://www.googletagmanager.com/gtm.js', { id: options?.id, l: options?.l }),`

Which makes it so we can easily change it like so:

```
googleTagManager: {
	id: 'GTM-123456',
	src: 'https://gtm.example.com/gtm.js'
},
```

Reyno-Skeps:
@ThiloHettmer Just submitted a PR.

blowsie:
Related https://github.com/nuxt/scripts/issues/397

blowsie:
This may help
https://github.com/nuxt/scripts/issues/434

PR #408, opened 2025-02-28
feat: added ability to change tagmanager and analytics script src

Reyno-Skeps:
### 🔗 Linked issue

Resolves https://github.com/nuxt/scripts/issues/404

### ❓ Type of change

### 📚 Description

Currently, there is no way to change the src of the tag manager script without losing the ability of the id being appended automatically. Changing the src is necessary for server side tagging to work.

This PR adds a new option to the tag manager (and analytics) options to change the src directly. This keeps the current functionality intact while allowing a custom src.

```
googleTagManager: {
	id: 'GTM-123456',
	src: 'https://gtm.example.com/gtm.js'
},
```


huang-julien (approved):

harlan-zw:
Hi, thanks for your work on this. I'm not sure if this is the best solution as there's already a property for overriding the source `scriptInput.src`.

You mention that you don't want to define the `id` twice, but this should not be required and is where the issue may lie instead.

```ts
// this should just work but I guess isn't
googleTagManager: {
	scriptInput: {
		src: 'https://gtm.example.com/gtm.js?id=GTM-123456',
	},
},
```

As Nuxt Scripts is focused on improving end-user performance, server-side tagging should be a first-class feature. I'd like to make sure that's the case but I haven't used it personally. 

As far as I can tell it's just configuring the `server_container_url`, but I guess from the original issue you're also proxying the script?

```ts
  gtag('config', 'TAG_ID', {
    server_container_url: 'https://analytics.example.com',
  });
```

PR #417, opened 2025-03-08
fix: resolve @unhead/vue from module and catch if fails

huang-julien:
### ❓ Type of change

### 📚 Description

We need to resolve @unhead/vue from the module location. 
A .catch has been added in case `readPackageJSON` wasn't able to find `@unhead/vue`.

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -117,7 +117,7 @@ export default defineNuxtModule<ModuleOptions>({
     }
     // couldn't be found for some reason, assume compatibility
     const { version: unheadVersion } = await readPackageJSON('@unhead/vue', {
-      from: import.meta.url,
+      from: nuxt.options.modulesDir,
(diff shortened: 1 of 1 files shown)


src/module.ts
```diff
@@ -116,7 +116,9 @@ export default defineNuxtModule<ModuleOptions>({
       return
     }
     // couldn't be found for some reason, assume compatibility
-    const { version: unheadVersion } = await readPackageJSON('@unhead/vue')
+    const { version: unheadVersion } = await readPackageJSON('@unhead/vue', {
+      from: import.meta.url,
```
danielroe: maybe use nuxt.options.modulesDir?

there are some examples in nuxt repo

huang-julien: 👍 thanks !

harlan-zw:
Thanks

PR #418, opened 2025-03-08
chore: add ohash v1 to docs deploy

danielroe:
### ❓ Type of change

aims to resolve docs deploy + unblock CI by adding ohash v1 manually - but only in docs (which ui-pro still depends on) - a similar strategy [worked for nuxt/fonts](https://github.com/nuxt/fonts/pull/558)

docs are still failing due to unenv hoisting


harlan-zw:
Much appreicated, thanks

PR #509, opened 2025-09-19
feat(stripe)!: Stripe basil

### LINKED ISSUE #508: Stripe script needs to be updated.
### 🐛 The bug

This specific script needs to be updated since according to stripe docs, integrations with that specific version are not recommended anymore:
[https://docs.stripe.com/sdks/stripejs-versioning](url)

This breaks many things when you follow the guide available in nuxt/scripts documentation, including typechecking since the api set in the script doesnt match the one in the recommended package.
`"@stripe/stripe-js"`

### 🛠️ To reproduce

https://stackblitz.com/edit/nuxt-starter-pmdyiz7a

### 🌈 Expected behavior

Types should match what the package exposes.

### ℹ️ Additional context

_No response_

gelanderosh:
### 🔗 Linked issue

"Resolves #508 " 

### ❓ Type of change:

- [X] 👌 Enhancement (improving an existing functionality)
- [X] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This pull request updates the url and type of the exposed stripe instance by the composable, when i followed the guide to implement stripe with nuxt/scripts i detected typescript errors that shouldnt be there, the reason was that the previous versions of stripe's library used to expose the constructor as Stripe, however in the newer versions the type is called StripeConstructor.

The reason im marking it as a breaking change is the fact that stripes api has changed considerably with the newer version, and updating nuxt/scripts would automatically update the library which can cause implementation problems because of the deprecated functions and such.

If this being a breaking change is a problem i can also implement a different script for the newer version of stripe. and maintain the previous as legacy.

### DIFF
--- a/src/runtime/registry/stripe.ts
+++ b/src/runtime/registry/stripe.ts
@@ -18,7 +18,7 @@ export function useScriptStripe<T extends StripeApi>(_options?: StripeInput) {
   return useRegistryScript<T, typeof StripeOptions>('stripe', options => ({
     scriptInput: {
       src: withQuery(
-        `https://js.stripe.com/v3/`,
+        `https://js.stripe.com/basil/stripe.js`,
         (typeof options?.advancedFraudSignals === 'boolean' && !options?.advancedFraudSignals) ? { advancedFraudSignals: false } : {},
       ),
       // opt-out of privacy defaults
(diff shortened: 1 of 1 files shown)


harlan-zw:
Awesome, thanks for the work :)

PR #521, opened 2025-09-29
fix: move @nuxt/kit into peerDependencies

### LINKED ISSUE #519: nodeReferences is undefined when running nuxt build (with v0.12)
### 🐛 The bug

Starting with v0.12.0 (also with v0.12.1), the build of my Nuxt 3 project doesn't work any more. I'm getting the following error:

```
 ERROR  Cannot read properties of undefined (reading 'push')
    at node_modules/@nuxt/scripts/node_modules/@nuxt/kit/dist/index.mjs:1286:22
    at node_modules/hookable/dist/index.mjs:48:66
    at node_modules/hookable/dist/index.mjs:48:56
    at async _generateTypes (node_modules/@nuxt/kit/dist/index.mjs:3528:3)
    at async writeTypes (node_modules/@nuxt/kit/dist/index.mjs:3563:37)
    at async Object.run (node_modules/@nuxt/cli/dist/chunks/prepare.mjs:49:5)
    at async runCommand (node_modules/citty/dist/index.mjs:316:16)
    at async runCommand (node_modules/citty/dist/index.mjs:307:11)
    at async runMain (node_modules/citty/dist/index.mjs:445:7) 
```

`node_modules/@nuxt/scripts/node_modules/@nuxt/kit/dist/index.mjs:1286:22` matches a `push` on `nodeReferences`.

I'm aware that the issues occurs in @nuxt/kit, but I wonder if the solution is about picking a correct version of it so that it works with Nuxt 3 or Nuxt 4.

In the StackBlitz reproduction project, the error is not the same. But if you download the project locally and try to run it, then you get the same error.

### 🛠️ To reproduce

https://stackblitz.com/edit/nuxt-starter-e3475ree

### 🌈 Expected behavior

No errors when building/running the project.

### ℹ️ Additional context

_No response_

huang-julien:
### 🔗 Linked issue

fix #519 

### ❓ Type of change

This PR moves @nuxt/kit to peerDependencies so users in Nuxt 3 wouldn't have @nuxt/kit 4 getting used due to the dependency graph of nuxt scripts

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
     "@types/google.maps": "^3.58.1",
     "@types/vimeo__player": "^2.18.3",
     "@types/youtube": "^0.1.0",
-    "@unhead/vue": "^2.0.3"
+    "@unhead/vue": "^2.0.3",
+    "@nuxt/kit": "^3.0.0 || ^4.0.0"
   },
   "peerDependenciesMeta": {
     "@stripe/stripe-js": {
@@ -98,7 +99,6 @@
     }
   },
   "dependencies": {
-    "@nuxt/kit": "^4.1.2",
     "@vueuse/core": "^13.9.0",
     "consola": "^3.4.2",
     "defu": "^6.1.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15,7 +15,7 @@ importers:
         specifier: ^2.6.2
         version: 2.6.2
       '@nuxt/kit':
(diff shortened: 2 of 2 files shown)


harlan-zw:
Thanks for the PR. 

The issue is that if we opt in to new @nuxt/kit features, the end user may not have them, so I'm not sure if merging this is the best option. It would be better to fix the original issue upstream.

Also, we should align to the https://github.com/nuxt/module-builder conventions

PR #623, opened 2026-03-04
fix: npm-mode registry scripts stored at $$scripts instead of $scripts

### LINKED ISSUE #622: npm-mode registry scripts stored at `$$scripts` instead of `$scripts`, making them inaccessible via the typed API
## Description

When using a registry script that uses `scriptMode: "npm"` internally (e.g. PostHog), the script stub is stored at `nuxtApp.$$scripts` (double `$`) instead of `nuxtApp.$scripts` (single `$`). This makes the stub completely inaccessible via the typed API and generated types, which correctly declare `NuxtApp.$scripts`.

The failure is **silent** — `useNuxtApp().$scripts?.posthog` returns `undefined` with no error, so any code that calls `.load()` or accesses `.proxy` on it simply does nothing.

## Reproduction

https://stackblitz.com/edit/nuxt-starter-punbbhjs

The repro page shows:
- `nuxtApp.$scripts?.posthog` → `undefined` (bug)
- `(nuxtApp as any).$$scripts?.posthog` → the actual stub (workaround)
- Two buttons: one calling `.load()` via each path — only the `$$scripts` path actually loads PostHog

```ts
const nuxtApp = useNuxtApp()

// What TypeScript says to use — always undefined for npm-mode scripts
nuxtApp.$scripts?.posthog        // undefined ❌

// Where the stub actually lives
;(nuxtApp as any).$$scripts?.posthog  // PosthogStub ✓
```

## Root Cause

CDN-mode scripts (`useScript()`) bypass `nuxtApp.provide()` entirely and write directly:

```js
// runtime/composables/useScript.js
nuxtApp.$scripts = nuxtApp.$scripts || reactive({})
nuxtApp.$scripts[id] = instance  // direct → $scripts ✓
```

npm-mode scripts go through the plugin's `provide` return value:

```js
// generated plugin (module.mjs)
return { provide: { $scripts: { posthog } } }
//                  ^ key already has '$' prefix
```

Nuxt's `applyPlugin` calls `nuxtApp.provide(key, value)` for each entry. `nuxtApp.provide` always prepends `$`:

```js
// nuxt/dist/app/nuxt.js
nuxtApp.provide = (name, value) => {
  const $name = '$' + name  // '$' + '$scripts' = '$$scripts'
  defineGetter(nuxtApp, $name, value)
}
```

Result:

```
plugin returns { provide: { $scripts: stub } }
  → nuxtApp.provide('$scripts', stub)
    → defineGetter(nuxtApp, '$$scripts', stub)   ← lands here
```

The types generated by the module declare `NuxtApp.$scripts`, which is right for CDN-mode but wrong for npm-mode at runtime.

## Proposed Fix

**Option A** — Remove the extra `$` from the plugin's provide key so Nuxt's mechanism produces `$scripts`:

```js
// Change in generated plugin:
return { provide: { scripts: { posthog } } }
//                  ^ no '$'; Nuxt adds it → '$scripts'
```

**Option B** — Bypass `provide` and assign directly, matching CDN-mode:

```js
nuxtApp.$scripts = nuxtApp.$scripts || reactive({})
nuxtApp.$scripts.posthog = posthogStub
```

Option A is a one-character fix.

## Workaround

```ts
const nuxtApp = useNuxtApp() as unknown as Record<string, unknown>
const stub = (nuxtApp['$$scripts'] as Record<string, unknown> | undefined)?.['posthog']
stub?.load()
```

## Environment

- `@nuxt/scripts`: `1.0.0-beta.7`
- `nuxt`: `^4.3.1`
- Affected: PostHog (and any other registry script using `scriptMode: "npm"`)
- Trigger mode where impact is most visible: `manual` (script never loads)

zizzfizzix:
## Summary

- Fixes a bug where npm-mode registry scripts (e.g. PostHog) were silently stored at `nuxtApp.$$scripts` instead of `nuxtApp.$scripts`, making them inaccessible via the typed API

## Root Cause

Nuxt's plugin `provide` mechanism automatically prepends `$` to every key. The generated plugin was returning `{ provide: { $scripts: ... } }` — with an explicit `$` already in the key — causing Nuxt to double it to `$$scripts`.

## Fix

Remove the extra `$` from the generated provide key (Option A as suggested by @harlan-zw):

```diff
- return { provide: { $scripts: { ... } } }
+ return { provide: { scripts: { ... } } }
```

Nuxt's `provide` then adds `$` → `$scripts` ✓

No type changes needed — the type declaration already correctly declares `NuxtApp.$scripts`.

Closes #622

### DIFF
--- a/src/templates.ts
+++ b/src/templates.ts
@@ -205,7 +205,7 @@ export function templatePlugin(config: Partial<ModuleOptions>, registry: Require
     `  parallel: true,`,
     `  setup() {`,
     ...inits.map(i => `    ${i}`),
-    `    return { provide: { $scripts: { ${[...Object.keys(config.globals || {}), ...resolvedRegistryKeys].join(', ')} } } }`,
+    `    return { provide: { scripts: { ${[...Object.keys(config.globals || {}), ...resolvedRegistryKeys].join(', ')} } } }`,
     `  }`,
     `})`,
   ].join('\n')


harlan-zw:
Looks like some CI issues but happy to merge and I'll fix these up myself. Thanks!

Issue #594, opened 2026-01-23
v1 Release


harlan-zw:
> Nuxt Scripts v1 is the first stable release.

Nuxt Scripts v1 is the first stable release, pushing the ecosystem forward for better **privacy** and **performance** for third-party scripts.

### 🔒 First-Party Mode: Privacy Focused Proxy

Every third-party script request exposes your users data to [fingerprinting](https://firstpartysimulator.net/learn).

Different providers are more intrusive, for example, the X Pixel accesses 9 browser fingerprinting APIs (including `navigator.getBattery()`), sets 5 tracking cookies, and makes requests to 3 separate domains. Microsoft Clarity reads 10 fingerprinting APIs across 3 domains.

First-party mode acts as a **reverse proxy**: scripts are bundled at build time and served from your domain, while runtime requests are securely forwarded through your server. Data sent to third-party servers gets anonymised: IPs (`180.233.124.74` -> `180.233.124.0`), browser versions (`Mozilla/5.0 (compatible; Chrome/120.0)`) and more. This is **auto-enabled** for all scripts that support it.

A side-effect is a performance boost from avoiding extra DNS lookups, fewer cookie banners, and reducing expensive fingerprinting queries. It also makes ad-blockers ineffective since requests appear same-origin.

See the [First-Party Mode Guide](https://scripts.nuxt.com/docs/guides/first-party) and [PR #577](https://github.com/nuxt/scripts/pull/577) for details.

### 🛠️ Rebuilt DevTools

The Nuxt DevTools panel has been rewritten around the v1 privacy and capability model. New views surface the things v1 does differently.

<img width="2764" height="1314" alt="Image" src="https://github.com/user-attachments/assets/47814398-b321-454f-895f-5d659d49a69f" />

### 🎉 Partytown Web Worker Support

Load third-party scripts off the main thread using [Partytown](https://partytown.builder.io/). Scripts run in a web worker, freeing the main thread for your app. Integrates directly with first-party mode. See [PR #576](https://github.com/nuxt/scripts/pull/576).

Set `partytown: true` per-script:

```ts
export default defineNuxtConfig({
  modules: ['@nuxtjs/partytown', '@nuxt/scripts'],
  scripts: {
    registry: {
      plausibleAnalytics: { domain: 'example.com', partytown: true },
      fathomAnalytics: { site: 'XXXXX', partytown: true },
      umamiAnalytics: { websiteId: 'xxx', partytown: true },
    }
  }
[...]

PR #699, opened 2026-04-10
feat(security): HMAC signing infrastructure for proxy endpoints

harlan-zw:
### 🔗 Linked issue

Related to #594

### ❓ Type of change

- [ ] 📖 Documentation
- [ ] 🐞 Bug fix
- [ ] 👌 Enhancement
- [x] ✨ New feature
- [ ] 🧹 Chore
- [ ] ⚠️ Breaking change

### 📚 Description

Proxy endpoints like Google Static Maps and Geocode inject a server-side API key and forward arbitrary requests to third-party services. Without signing, anyone can call these endpoints directly and burn the site owner's API quota. This PR adds the signing infrastructure so individual endpoints can opt in.

**What's included:**

- **Sign/verify primitives** (`runtime/server/utils/sign.ts`): canonical query serialization (sorted keys, sig-stripped, URL-encoded), HMAC-SHA256 with 16-char hex signatures, constant-time comparison
- **`withSigning` wrapper** (`runtime/server/utils/withSigning.ts`): opt-in middleware that returns 403 on missing/invalid signatures, 500 if secret not configured
- **Secret management** (`module.ts`): `NUXT_SCRIPTS_PROXY_SECRET` env var, `scripts.security.secret` config, dev auto-generates to `.env`, prod throws if missing when signed endpoints are registered
- **`/_scripts/sign` endpoint** (`runtime/server/sign-proxy.ts`): for reactive client-side URL updates (e.g. Google Static Maps recomputing size on mount); same-origin check + per-IP rate limiting (60/min)
- **CLI** (`src/cli.ts` + `bin/cli.mjs`): `npx @nuxt/scripts generate-secret` for explicit secret management
- **`requiresSigning` flag** on `RegistryScriptServerHandler` so endpoints can declare they need signing
- **23 unit tests** covering canonicalization, roundtrip verification, tamper rejection, cross-endpoint isolation

**Not included (follow-up PRs):**
- Wiring `requiresSigning: true` on actual handlers (google-static-maps, geocode, gravatar, embed image proxies)
- Updating components to call `buildSignedProxyUrl` at SSR time
- Hardening `createImageProxyHandler` (streaming size limit, DNS + private-IP check)
- Body-size cap on `/_scripts/p/**` analytics reverse proxy


harlan-zw:
### Code review

Found 1 issue:

1. Em dashes used throughout new files, violating the writing style rule in `.claude/CLAUDE.md`: "Never use em dashes (—) or hyphens as dashes (-) in writing. Restructure sentences using commas, semicolons, colons, or separate sentences instead." Approximately 10 instances across `cli.ts`, `module.ts`, `sign-proxy.ts`, and `withSigning.ts`, in both code comments and CLI output strings.

https://github.com/nuxt/scripts/blob/e8434f1d4cfc0948730ca96f2375a53e9b5886d6/packages/script/src/cli.ts#L3-L7

https://github.com/nuxt/scripts/blob/e8434f1d4cfc0948730ca96f2375a53e9b5886d6/packages/script/src/runtime/server/sign-proxy.ts#L5-L8

https://github.com/nuxt/scripts/blob/e8434f1d4cfc0948730ca96f2375a53e9b5886d6/packages/script/src/module.ts#L147-L149

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

harlan-zw:
### Code review

Found 2 issues:

1. `module.ts` fatal throw breaks CI and `nuxt prepare`. The guard `anyHandlerRequiresSigning && !proxySecretResolved?.secret && !nuxt.options.dev` fires during `nuxt prepare` (which runs in non-dev mode) when no `NUXT_SCRIPTS_PROXY_SECRET` is set. Since all 10 registry handlers now have `requiresSigning: true`, any fixture or project enabling these scripts fails at prepare time. This contradicts the `withSigning` pass-through behavior (which is designed to be non-breaking). The throw should be gated on whether the secret was explicitly requested (e.g. `config.security?.secret` was set), or downgraded to a warning.

https://github.com/nuxt/scripts/blob/deacf59ce9a16ff9296562391877a1a98c5cef6c/packages/script/src/module.ts#L876-L884

2. `resolveProxySecret` runs unconditionally before `anyHandlerRequiresSigning` is determined. In dev, it auto-generates a secret and writes `NUXT_SCRIPTS_PROXY_SECRET` to `.env` even when zero proxy endpoints are enabled. Only 5 of 38 registry scripts have `serverHandlers`; users selecting only client-side scripts (analytics, tracking) get an unexpected env var in their `.env`. The call should move after the handler registration loop and only execute when `anyHandlerRequiresSigning` is true.

https://github.com/nuxt/scripts/blob/deacf59ce9a16ff9296562391877a1a98c5cef6c/packages/script/src/module.ts#L486-L498

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #712, opened 2026-04-14
feat: vendor-native consent controls for consent-aware scripts

### LINKED ISSUE #711: Tiktok Pixel consent mode
### 🆒 Your use case

Tiktok Pixel supports consent mode where consent can be held and explicitly granted/revoked at a later time. While consent is being held, events will be queued until the user grants or revokes consent, in which case events will either be sent or held in queue. Documentation can be found here: https://business-api.tiktok.com/portal/docs/pixel-cookie-consent-mode/v1.3 

### 🆕 The solution you'd like

This is somewhat similar to Google Analytics and Bing UET with their `onBefore<script>Start`, but not quite as the methods need to be added to the `ttq.methods` array. This should perhaps be an option like:
```js
const script = useScriptTikTokPixel({
  id: '<pixel id>',
  defaultConsent: 'opt-in', // options could be 'opt-out' | 'opt-in'
})
```
Alternatives would also be welcome.

Alternative solution would be to use `useScript` and implement it manually.

### ℹ️ Additional info

It's a bit unclear how to get access to the Tiktok methods from the current registry script as it also includes methods like `enableCookie`/`disableCookie`. It would be nice to have those implemented in some way too, but mostly as a nice to have.

harlan-zw:
### 🔗 Linked issue

Resolves #711
Related to #243, #710

### ❓ Type of change

- [ ] 📖 Documentation
- [ ] 🐞 Bug fix
- [ ] 👌 Enhancement
- [x] ✨ New feature
- [ ] 🧹 Chore
- [ ] ⚠️ Breaking change

### 📚 Description

Adds first-class consent controls to every consent-aware registry script, so cookie banners can wire up without hand-rolling `onBefore*Start` callbacks. Each script now exposes a vendor-native `consent` object on the returned instance plus a typed `defaultConsent` option applied inside `clientInit` before the first tracking call.

The API is intentionally not unified: each vendor speaks its own dialect (GCMv2 for Google/Bing, `grant`/`revoke` for Meta, `grantConsent`/`revokeConsent`/`holdConsent` for TikTok, `opt_in`/`opt_out` for Mixpanel and PostHog, `requireConsent`/`setConsentGiven` for Matomo, `clarity('consent', ...)` for Clarity). Matching the vendor shape means the docs, types, and vendor guides line up one-to-one.

### What shipped

**`defaultConsent` option** — applied before the vendor init / first tracking call so the initial page view honors the user's stored decision.

| Script | `defaultConsent` shape |
|---|---|
| Google Analytics | `Partial<ConsentState>` (GCMv2) |
| Google Tag Manager | `Partial<ConsentState>` (GCMv2) |
| Bing UET | `{ ad_storage }` |
| Meta Pixel | `'granted' \| 'denied'` |
| TikTok Pixel | `'granted' \| 'denied' \| 'hold'` |
| Matomo | `'required' \| 'given' \| 'not-required'` |
| Mixpanel | `'opt-in' \| 'opt-out'` |
| PostHog | `'opt-in' \| 'opt-out'` |
| Clarity | `boolean \| Record<string, string>` |

**Vendor-native `consent` object** — exposed on the `UseScriptContext` for post-consent updates:

```ts
const gtm = useScriptGoogleTagManager({
  id: 'GTM-XXX',
  defaultConsent: { ad_storage: 'denied', analytics_storage: 'denied' },
})
// later, after the user accepts
gtm.consent.update({ ad_storage: 'granted', analytics_storage: 'granted' })

const ttq = useScriptTikTokPixel({ id: '...', defaultConsent: 'hold' })
ttq.consent.grant()   // or .revoke() / .hold()

const meta = useScriptMetaPixel({ id: '...', defaultConsent: 'denied' })
meta.consent.grant()  // or .revoke()
```

Escape hatches (`onBeforeGtagStart`, `onBeforeUetStart`, etc.) retained.

### Notes

- GCMv2 `ConsentState` is exported from `#nuxt-scripts/types` for typed banner code.
- Matomo's `consent.grant()` / `revoke()` are no-ops (with dev warning) unless `defaultConsent: 'required' \| 'given'` was set at registration — mirrors Matomo's own requirement.
- TikTok `holdConsent` / `grantConsent` / `revokeConsent` stubs are queued before `events.js` binds, so pre-load calls work.
- Reddit Pixel, X Pixel, Snapchat Pixel skipped — no documented runtime consent API to adapt to.

Full guide updated in `docs/content/docs/1.guides/3.consent.md` with OneTrust / Cookiebot recipes and the vendor mapping table.


harlan-zw:
### Code review

Found 2 issues:

1. Adapter registration passes the raw `ScriptInstance` as the proxy, not `instance.proxy`. Every `consentAdapter` accesses vendor APIs off the proxy (e.g. `proxy.ttq.grantConsent()`, `proxy.fbq(...)`, `proxy.uetq.push(...)`), so `applyDefault` / `applyUpdate` will throw a `TypeError` at runtime. The granular fan-out for every consent-aware script is broken.

https://github.com/nuxt/scripts/blob/3b68b6c6ba2be90539f410f9029a1e23713c838c/packages/script/src/runtime/composables/useScript.ts#L286-L298

2. PostHog declares a `consentAdapter` in the registry, but because PostHog uses `scriptMode: 'npm'`, `useRegistryScript` early-returns into `createNpmScriptStub` before the consent wiring in `useScript.ts` ever runs. The adapter is never subscribed, so granular GCMv2 fan-out silently does nothing for PostHog even though the consent guide's vendor table promises it.

https://github.com/nuxt/scripts/blob/3b68b6c6ba2be90539f410f9029a1e23713c838c/packages/script/src/runtime/utils.ts#L68-L78

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #730, opened 2026-04-28
fix(google-maps): persist user pan and re-emit ready on color-mode re-init

harlan-zw:
### 🔗 Linked issue

Related to #727

### ❓ Type of change

- [ ] 📖 Documentation
- [x] 🐞 Bug fix
- [ ] 👌 Enhancement
- [ ] ✨ New feature
- [ ] 🧹 Chore
- [ ] ⚠️ Breaking change

### 📚 Description

Two regressions surfaced after the cloud-styled mapId fix in #727:

1. **User pan discarded on color-mode toggle.** The recreate watcher captured the live center before teardown, but the standalone center watcher then re-fired against the new map instance and called `setCenter(propsInitialCenter)`. Persist the captured center into `centerOverride` so the recomputed `options.value.center` matches the new map's actual center; the watcher's lat/lng comparison guard then short-circuits.

2. **Imperative resources silently dropped.** Pins (or any objects) created off the exposed `map` ref outside declarative children disappeared after re-init with no signal to re-attach them. Re-emit `ready` after constructing the new Map instance so consumers can re-bind, and document the contract in the map-styling guide.

Adds regression tests covering both behaviours.

### DIFF
--- a/docs/content/scripts/google-maps/1.guides/2.map-styling.md
+++ b/docs/content/scripts/google-maps/1.guides/2.map-styling.md
@@ -97,7 +97,9 @@ If you set up a single Map ID in Google Cloud Console with both Light and Dark c
 ```
 
 ::callout{color="amber"}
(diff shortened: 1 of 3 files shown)


harlan-zw:
Duplicate of #731 (same fix, less complete). Closing in favor of #731.

PR #739, opened 2026-05-02
fix(google-analytics, google-tag-manager): per-region consent defaults

zizzfizzix:
### 🔗 Linked issue

None — searched open and closed issues; no existing tracker for this.

### ❓ Type of change

- [ ] 📖 Documentation
- [ ] 🐞 Bug fix
- [x] 👌 Enhancement
- [x] ✨ New feature
- [ ] 🧹 Chore
- [ ] ⚠️ Breaking change

### 📚 Description

Extends the `defaultConsent` API introduced in #712 so that `useScriptGoogleAnalytics` and `useScriptGoogleTagManager` accept an **array** of GCMv2 states, not just a single object. Each entry fires its own `gtag('consent', 'default', entry)` call, matching Google's documented multi-call pattern for [region-specific consent](https://developers.google.com/tag-platform/security/guides/consent?consentmode=advanced#region-specific-behavior).

```ts
useScriptGoogleTagManager({
  id: 'GTM-XXXX',
  defaultConsent: [
    {
      ad_storage: 'denied',
      analytics_storage: 'denied',
      region: ['ES', 'US-AK', /* … EEA codes */],
      wait_for_update: 500,
    },
    { ad_storage: 'granted', analytics_storage: 'granted' }, // unscoped global fallback
  ],
})
```

Backwards compatible: passing a single object behaves exactly as before. Single-entry arrays are observationally equivalent to the bare object (locked by a test). Empty arrays are a documented no-op.

### What changed

- **GA + GTM \`defaultConsent\`** widened to \`union(gcmConsentState, array(gcmConsentState))\`. Runtime normalizes to an array and iterates.
- **GTM schema strictening (drive-by):** the previous \`record(string, string|number)\` was looser than GA's already-canonical \`gcmConsentState\` and didn't model \`region\` as an array. GTM now uses the same canonical type as GA. Schema validation only runs in dev, so the runtime impact is dev-mode warnings only — but anyone passing arbitrary unknown keys in \`defaultConsent\` today will see a new dev validation error.
- **\`ConsentState\` interface (drive-by):** gains \`region: string[]\` and \`wait_for_update: number\` to match what the runtime schema already accepts.
- **GTM \`@see\` URL alignment (drive-by):** both schemas now point at the canonical \`?consentmode=advanced#region-specific-behavior\` Google docs anchor.
- **\`registry-types.json\`** regenerated via \`pnpm generate:types\` to reflect the schema changes.
- **New tests** in \`test/nuxt-runtime/consent-default.nuxt.test.ts\`: array ordering vs. \`gtm.js\` / \`gtag('js', …)\`, single-entry-array == bare-object equivalence, empty-array no-op. Three for GTM, three for GA. Also closes a pre-existing gap where GA had no consent-ordering test.
- **Docs** updated on both GA and GTM script pages with a \"Per-region defaults\" section + worked example. The GA page's previously-h3 \`Consent Mode\` section was promoted to h2 to match the cross-page convention; \`Customer/Consumer ID Tracking\` was promoted along with it to keep the heading hierarchy honest.
- **New example** \`examples/regional-consent/\` showing EEA-denied + global-granted defaults.

### Out of scope

- **Bing UET.** Microsoft's UET consent mode docs do not support a \`region\` field; supporting multi-call there without a vendor spec to back it would mislead users. Easy follow-up if Microsoft documents region support later.
- All other consent-aware scripts (Matomo, Clarity, PostHog, Mixpanel, TikTok Pixel, Meta Pixel) — they have non-GCMv2 consent shapes; per-region work for those would need separate vendor research.


harlan-zw:
Looks good, thanks! Are you able to take a look at coderabbit comments and fix if relevant + resolve them.

zizzfizzix:
Addressed CodeRabbit's review in `a8bef9a`:

1. **GTM `defaultConsent` before `onBeforeGtmStart`** — applied. Real privacy-correctness fix; brings GTM in line with GA's existing ordering. Custom events pushed inside `onBeforeGtmStart` now honour the configured consent state.
2. **Equivalence-test slices include the `gtm.js` / `gtag('js', new Date())` timestamp** — fixed. Dropped the `+ 1` from both slices so the timestamps are no longer compared. Ordering relative to those entries is locked by the sibling tests already, so coverage is unchanged.
3. **GTM `defaultConsent` JSDoc ends with dangling `See:`** — fixed. Rewrote the description so it ends with the parenthetical about precedence; the `@see` lines are still attached but no longer leak a trailing `See:` into `registry-types.json`. Regenerated `registry-types.json`.
4. **Nitpick: pin `@nuxt/scripts` instead of `latest` in `examples/regional-consent/package.json`** — leaving as-is. This matches the existing pattern in `examples/cookie-consent/package.json`, `examples/granular-consent/package.json`, etc. Pinning across the `examples/*` tree is a separate decision worth making consistently or not at all; happy to do it as a follow-up if the maintainer wants.

@coderabbitai please re-review.

PR #754, opened 2026-05-07
chore: move top-level Google Maps `center`/`zoom` examples under `map-options`

DamianGlowala:
### 📚 Description

Addresses deprecation of top-level `center`/`zoom` Google Maps opts by updating docs/playground examples.


harlan-zw:
Thank you :heart:

Issue #761, opened 2026-05-10
Warn "missing required field id" for registry scripts when id is only set via runtimeConfig / NUXT_PUBLIC_SCRIPTS_*


tsukiyama-3:
### 🐛 The bug

When the pixel / tag id is set only via runtimeConfig.public.scripts and/or NUXT_PUBLIC_SCRIPTS_* (no id in scripts.registry), the dev server can still log:

```
[nuxt-scripts] registry.<key>: missing required field 'id'. The script infrastructure is registered but will not function without it.
```

By then runtimeConfig.public.scripts already has id after the module merge, so the warning is wrong and clashes with the documented env / runtimeConfig flow.

Docs: [Nuxt Config — Environment Variables](https://scripts.nuxt.com/docs/api/nuxt-config#environment-variables)

### 🛠️ To reproduce

https://stackblitz.com/edit/github-gzbuzr-ghdbaa84?file=app%2Fapp.vue,app%2Fpages%2Findex.vue,nuxt.config.ts,package.json

### 🌈 Expected behavior

No WARN when the effective config (after merging process.env / registryWithDefaults into runtimeConfig.public.scripts) satisfies the schema’s required fields.

If id is truly missing everywhere, the WARN should still appear.

PR #762, opened 2026-05-10
fix: validate required registry fields after runtime merge

tsukiyama-3:
### 🔗 Linked issue

https://github.com/nuxt/scripts/issues/761

### 📚 Description

Required-field validation during module setup compared the Valibot schema to the normalized `[input]` from `scripts.registry` only.

IDs and similar fields are often supplied only through `runtimeConfig.public.scripts` and/or `NUXT_PUBLIC_SCRIPTS_*`, merged into `public.scripts` earlier in this module (`registryWithDefaults` + `defu`). After that merge, the effective config already includes `id`, but the check still complained.

The WARN now inspects each script’s merged object in `runtimeConfig.public.scripts`, dropping `scriptOptions` keys that aren’t part of script input schema. Behaviour at runtime is unchanged; this only aligns dev warnings with documented env/runtime configuration.


harlan-zw:
Thanks!

PR #782, opened 2026-05-18
feat: SpeedCurve LUX

zizzfizzix:
### 🔗 Linked issue

Related to #538

### ❓ Type of change

- [ ] 📖 Documentation
- [ ] 🐞 Bug fix
- [ ] 👌 Enhancement
- [x] ✨ New feature
- [ ] 🧹 Chore
- [ ] ⚠️ Breaking change

### 📚 Description

Adds SpeedCurve LUX to the script registry with `useScriptSpeedCurve()`, covering the full LUX SDK: automatic page-view tracking, SPA navigation support, and manual mode. Registry entry uses a 13-field Valibot schema for LUX configuration (`samplerate`, `sendBeaconOnPageHidden`, `trackErrors`, `maxErrors`, `minMeasureTime`, `maxMeasureTime`, `newBeaconOnPageShow`, `trackHiddenPages`, `cookieDomain`, plus SPA-specific `spaMode`, `autoTrackSpaNavigations`, `label`).

**Design notes**

- `afterNextPaint(fn)` schedules `markLoadTime` after the next `requestAnimationFrame` flush so it runs after LUX's paint capture window closes.
- SPA mode: with `autoTrackSpaNavigations` enabled, `vue-router` `beforeEach` calls `lux.startSoftNavigation()` and sets the label; the `page:finish` Nuxt hook calls `afterNextPaint(() => lux.markLoadTime())` to seal the beacon after render; `afterEach` seals cancelled-navigation phantom beacons with `markLoadTime` + `addData('luxNavFailed', '1')`.
- `label` accepts `string | ((to: RouteLocationNormalized) => string) | false`; functions resolve per-navigation, `false` skips it entirely, default falls back to `String(to.name ?? to.path)`.
- HMR guard (`luxWired` module flag) prevents double-wiring the router hook on Vite hot reloads.

Ships logo (light/dark SVG), playground pages for each SPA mode (`/third-parties/speed-curve/spa-auto`, `no-spa`, `manual-spa`), docs at `docs/scripts/speedcurve.md`, plus 26 unit tests and 11 Playwright e2e tests covering primer injection, config application, auto-tracker, `afterNextPaint`, and SPA scenarios (default/custom/disabled label, back navigation, cancelled navigation, manual mode).


packages/script/src/runtime/registry/speedcurve.ts
```diff
@@ -0,0 +1,135 @@
...
+    if (v === undefined)
+      continue
+    // label may be a function or false (handled by installAutoTracker); only pass strings to LUX directly
+    if (k === 'label' && typeof v !== 'string')
+      continue
+    lux[k as string] = v
+  }
+}
+
+export function installAutoTracker(options?: SpeedCurveInput): void {
```
harlan-zw, line 82: We should use https://scripts.nuxt.com/docs/api/use-script-event-page ?

packages/script/src/runtime/registry/schemas.ts
```diff
@@ -1007,6 +1007,75 @@ export const InitObjectPropertiesSchema = object({
   age: optional(string()),
 })
 
+export const SpeedCurveOptions = object({
```
harlan-zw, line 1010: We should be explicit in whats speedsurve api options vs what we're baking in

packages/script/src/runtime/registry/speedcurve.ts
```diff
@@ -0,0 +1,135 @@
+import type { LuxGlobal, UserConfig } from '@speedcurve/lux'
+import type { RouteLocationNormalized } from 'vue-router'
+import type { RegistryScriptInput, UseScriptContext } from '#nuxt-scripts/types'
+import luxSnippetSource from '@speedcurve/lux/dist/lux-snippet.js?raw'
```
harlan-zw, line 4: this needs to be a peer dep?

harlan-zw:
Thanks for the PR, for your descriptions, can you please not use ai generated summaries - just write it. Left a couple of blocking comments.

packages/script/src/runtime/registry/speedcurve.ts
zizzfizzix: `useScriptEventPage` fires after `dom:rendered` (or a 100ms timeout), both of which happen before `afterNextPaint` would be called so nesting them would add up to 100ms of unnecessary delay before `markLoadTime()`, which would skew SpeedCurve's measurements.
We could theoretically extend `useScriptEventPage` with `{ onBeforeNavigate, onNavigationFailed }` to consolidate the router guards, but then with `onChange` unused it feels a bit pointless.
What do you think?

packages/script/src/runtime/registry/schemas.ts
zizzfizzix: What would you prefer, JSDoc section comments within the flat object to label the composable-level keys (id, spaMode, autoTrackSpaNavigations) vs the passthrough LUX UserConfig keys, or a split (e.g. a nested lux: {} object for the UserConfig portion), or something else?

packages/script/src/runtime/registry/speedcurve.ts
harlan-zw: Good point, this looks good then

packages/script/src/runtime/registry/speedcurve.ts
zizzfizzix: Good catch, I see 2 options:
1. Vendor the snippet: pull the primer out into a speedcurve-primer.ts with a plain string const. The @speedcurve/lux dep becomes types-only like other scripts.
**This would need a manual update** when SpeedCurve eventually bumps the primer.
2. Inline via rollup: a plugin in build.config.ts that handles ?raw at nuxt-scripts build time. The advantage is it auto-tracks upstream.

What do you think?

harlan-zw:
Thanks!

zizzfizzix:
@harlan-zw nice!
Two of your comments were unresolved, if you can give some guidance I'm happy to address them in a follow-up PR.

harlan-zw:
Uh i'll sort it out :+1:  trying to land a release, i'll let you know once its merged and you can PR any changes needed

PR #787, opened 2026-05-21
fix(security): add pageToken opt-out to keep SSR payload deterministic

### LINKED ISSUE #783: Nuxt Scripts injects proxy token into payload even if not used
### 🐛 The bug

Based on [this](https://github.com/nuxt/scripts/blob/main/packages/script/src/runtime/plugins/proxy-token.server.ts), the only way to not have the proxy token shoved into the Nuxt payload is to not have the secret set, but because it is auto-generated in dev, I can't actually test the payload properly.

The actual issue is that I'd like to calculate an `etag` for the response via a Nitro plugin, but because the proxy token is being included in the payload for every request, each hash is different. I had assumed that just setting `proxy: false` for the one script I was including would turn off this inclusion, but it does not.

### 🛠️ To reproduce

https://stackblitz.com/edit/nuxt-starter-rxyuphdo?file=nuxt.config.ts

### 🌈 Expected behavior

Don't include the proxy token if it isn't going to be used (or warn that the environment variable isn't set). If it needs to be a flag set at a higher level, that's fine with me.

Obviously, just not including the `NUXT_SCRIPTS_PROXY_SECRET` for production builds should work fine without any changes, but it is going to yell at me and still not work properly for dev.

### ℹ️ Additional context

_No response_

harlan-zw:
### 🔗 Linked issue

Resolves #783

### ❓ Type of change

- [ ] 📖 Documentation
- [x] 🐞 Bug fix
- [ ] 👌 Enhancement
- [ ] ✨ New feature
- [ ] 🧹 Chore
- [ ] ⚠️ Breaking change

### 📚 Description

The per-request proxy page token is embedded in the SSR payload whenever a signing-required proxy handler is enabled. Being timestamp-based, it makes the payload differ on every request, which breaks response `etag` hashing.

Adds `security.pageToken` (default `true`). When `false`, the proxy-token server plugin is not registered, so no token is generated and the payload stays deterministic. Client-driven proxy URLs must then be pre-signed.

**Default behaviour is unchanged** — this is a pure opt-out, so it carries no regression for SSG, embeds, or client-side navigation. The issue author noted a higher-level flag would be acceptable.

Verified against `test/fixtures/issue-783`: with `pageToken: false`, `useScriptProxyToken` resolves to null and the SSR payload is byte-identical across requests.

### DIFF
--- a/docs/content/docs/1.guides/2.first-party.md
+++ b/docs/content/docs/1.guides/2.first-party.md
@@ -280,6 +280,10 @@ export default defineNuxtConfig({
       // Auto-generate and persist a secret to .env in dev mode.
       // Set to false to disable.
       autoGenerateSecret: true,
+      // Emit a per-request page token into the SSR payload.
+      // Set false for a deterministic payload (e.g. response etag hashing);
+      // client-driven proxy URLs must then be pre-signed.
+      pageToken: true,
     }
   }
 })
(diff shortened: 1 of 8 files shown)


harlan-zw:
Superseded by #788, which removes the proxy URL signing subsystem entirely (the page token was only one symptom of a mechanism that also breaks under caching/SSG and could not protect the one endpoint that mattered).

PR #795, opened 2026-05-26
fix(deps): add support for unhead v3

danielroe:
### 🔗 Linked issue

closes https://github.com/nuxt/scripts/pull/703

### 📚 Description

this is an initial stab at making nuxt/scripts pass in ecosystem-ci with nuxt v4.5 (which contains unhead v3)

I've made a few guesses but feel free to scrap this!

### DIFF
--- a/packages/script/src/runtime/components/GoogleMaps/ScriptGoogleMapsStaticMap.vue
+++ b/packages/script/src/runtime/components/GoogleMaps/ScriptGoogleMapsStaticMap.vue
@@ -254,10 +254,10 @@ if (import.meta.server) {
   useHead({
     link: [
       {
-        rel: props.loading === 'eager' ? 'preconnect' : 'dns-prefetch',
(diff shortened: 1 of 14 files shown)


packages/script/src/runtime/components/GoogleMaps/ScriptGoogleMapsStaticMap.vue
```diff
@@ -254,7 +254,7 @@ if (import.meta.server) {
   useHead({
     link: [
       {
-        rel: props.loading === 'eager' ? 'preconnect' : 'dns-prefetch',
+        rel: (props.loading === 'eager' ? 'preconnect' : 'dns-prefetch') as 'preconnect',
```
danielroe: is this correct? do we not support `dns-prefetch` as a rel value

harlan-zw: the type issue was from the .filter() / undefined href, these are both properly unionised

PR #815, opened 2026-06-12
fix(umami): prevent SSR crash by guarding window access in use()

DrMartinGonzo:
Not very familiar with the codebase, but here goes, it fixed my problems after many hours of not finding useful logs :)

### 📚 Description

This PR fixes a fatal SSR crash (window is not defined) that occurs when using the Umami Analytics script registry in universal rendering mode.

Currently, the use() method in packages/script/src/runtime/registry/umami-analytics.ts eagerly attempts to return window.umami without an environment check. If this method is called within a component's setup phase, it crashes the Node.js/Nitro server during the initial render.

I simply added an import.meta.client guard to the use() method in the Umami registry.

During server-side execution, it now safely bypasses the window access, preventing the crash while preserving the expected behavior on the client.

### DIFF
--- a/packages/script/src/runtime/registry/umami-analytics.ts
+++ b/packages/script/src/runtime/registry/umami-analytics.ts
@@ -48,7 +48,10 @@ export function useScriptUmamiAnalytics<T extends UmamiAnalyticsApi>(_options?: 
       schema: import.meta.dev ? UmamiAnalyticsOptions : undefined,
       scriptOptions: {
         use() {
-          return window.umami
+          if (import.meta.client) {
+            return window.umami
+          }
+          return undefined
         },
       },
     }


harlan-zw:
.use() shouldn't be called server-side, do you have any reproduction code? there's possibly a root issue to fix

DrMartinGonzo:
I was suspecting that. Thank you for your answer Harlan.
I will try to strip my project to the bare minimum and see if I can reproduce. Will share once its done.

DrMartinGonzo:
Thank you for putting me in the right direction.
After some digging, I've been able to set a minimal repro.
I found out a very particular setup causes the problem.

Having "@nuxtjs/seo" and "@nuxt/scripts" enabled.
Using experimental `viteEnvironmentApi: true` and `ssr: true` in nuxt.config.ts.

And for some reason, only fails when building in a docker container based on node-alpine, ok outside. (node and pnpm both being the same version as on my computer). I don't know how to debug this further.

Since this is related to an experimental nuxt feature, I guess we can close this.

harlan-zw:
Thanks for the details. I've pushed a fix here https://github.com/nuxt/scripts/pull/820

PR #855, opened 2026-08-07
fix(gtm): don't declare `dataLayer` on the global `Window`

### LINKED ISSUE #852: GTM registry's global `Window` augmentation is irreconcilable with @gtm-support/core — unsuppressable TS2430 under TypeScript 7
### 🐛 The bug

`dist/runtime/registry/google-tag-manager.d.ts` augments the global `Window` unconditionally:

```ts
export interface GoogleTagManagerApi {
  google_tag_manager: GoogleTagManagerInstance
  dataLayer: DataLayer & { push: DataLayerPush }   // ← required
}

declare global {
  interface Window extends GoogleTagManagerApi {}
}
```

`@gtm-support/core` (the engine behind [`@gtm-support/vue-gtm`](https://github.com/gtm-support/vue-gtm), a widely used Vue/Nuxt GTM integration) declares the same member as **optional**:

```ts
// @gtm-support/core@3.0.1 lib/index.d.ts
declare global {
  interface Window {
    dataLayer?: DataLayerObject[]
  }
}
```

Interface declarations merge, so `Window.dataLayer` becomes `DataLayerObject[] | undefined`, which then fails the `extends GoogleTagManagerApi` clause this module itself added:

```
error TS2430: Interface 'Window' incorrectly extends interface 'GoogleTagManagerApi'.
  Types of property 'dataLayer' are incompatible.
    Type 'DataLayerObject[] | undefined' is not assignable to type 'DataLayer & { push: DataLayerPush; }'.
      Type 'undefined' is not assignable to type 'DataLayer & { push: DataLayerPush; }'.
        Type 'undefined' is not assignable to type 'DataLayer'.
```

Three things make this hard to live with:

1. **It fires even if you never touch GTM via this module.** We use `useScriptGoogleAnalytics` and have never called `useScriptGoogleTagManager` — the augmentation ships regardless, because the registry composable's auto-import pulls the `.d.ts` into the program.
2. **It is unsuppressable from the consumer side.** The error is reported at *our* `declare global { interface Window { … } }` sites, one per augmentation block, not at either dependency. With `skipLibCheck: true` a dependency-vs-dependency conflict is silently ignored; it only materialises because we also augment `Window`.
3. **`tsc` doesn't warn you — it just picks a winner.** `tsc@5.9.3` exits 0 here, because it does not verify merged interfaces against their `extends` bases. But the type it resolves is the *optional* one, so `GoogleTagManagerApi`'s contract quietly loses:
   ```ts
   window.dataLayer.push({ event: 'x' })
   //     ~~~~~~~~~ TS18048: 'window.dataLayer' is possibly 'undefined'
   ```
   TypeScript 7 / `tsgo` **does** perform that check, which is how we found this. So this will surface for many more people as TS 7 adoption grows — it is not a `tsgo` bug.

### 🛠️ Minimal reproduction

Three files, `strict: true`, `skipLibCheck: true`, `moduleResolution: "Bundler"`, `lib: ["ESNext","DOM"]`:

```ts
// upstream.d.ts — mirrors this module's GTM registry
interface DataLayerObject { [key: string]: any }
type DataLayer = Array<DataLayerObject>
type DataLayerPush = (...args: any[]) => number
interface GoogleTagManagerInstance { dataLayer: { gtmDom: boolean, [k: string]: unknown }, [id: string]: any }
interface GoogleTagManagerApi {
  google_tag_manager: GoogleTagManagerInstance
  dataLayer: DataLayer & { push: DataLayerPush }
}
declare global { interface Window extends GoogleTagManagerApi {} }
export {}
```

```ts
// gtmsupport.d.ts — mirrors @gtm-support/core
interface GtmDataLayerObject extends Record<string, any> { event?: string }
declare global { interface Window { dataLayer?: GtmDataLayerObject[] } }
export {}
```

```ts
// ours.ts — any first-party Window augmentation
declare global { interface Window { GTest: string } }
export {}
```

- `tsgo` / `tsc@7.0.2` → `TS2430` at `ours.ts`
- `tsc@5.9.3` → exits 0 (and types `window.dataLayer` as possibly-undefined)
- Delete `ours.ts` → `tsgo` also exits 0, because `skipLibCheck` hides the lib-vs-lib conflict

### 💡 Suggested fix

Make `dataLayer` optional in `GoogleTagManagerApi`:

```diff
 export interface GoogleTagManagerApi {
   google_tag_manager: GoogleTagManagerInstance
-  dataLayer: DataLayer & { push: DataLayerPush }
+  dataLayer?: DataLayer & { push: DataLayerPush }
 }
```

This is arguably more accurate independently of the conflict: `window.dataLayer` is created by the GTM snippet at load time, so before the script loads it genuinely is `undefined`. It would also make the two declarations compatible, since optional-vs-optional merges cleanly.

Alternatives, if changing the public type is undesirable:

- Keep `GoogleTagManagerApi` as-is but drop the `declare global { interface Window extends GoogleTagManagerApi {} }` block. The type is already exported and used as `useScriptGoogleTagManager<T extends GoogleTagManagerApi>`, so the global augmentation is what creates the collision without being needed for the composable's own typing.
- Gate the augmentation behind the registry actually being configured, so projects that don't use GTM don't inherit it. (Related: #135, where registry types for unused scripts broke a consumer's typecheck.)

### ℹ️ Additional context

- `@nuxt/scripts@1.3.2`, `@gtm-support/vue-gtm@3.2.0` → `@gtm-support/core@3.0.1`, Nuxt 4.4.8, `vue-tsc` 3.x.
- Present in `0.13.2` too, with a byte-identical declaration — this is long-standing, not a recent regression.
- The required `dataLayer` was added in #388 ("fix(googleTagManager): add missing dataLayer type"), so I assume it is deliberate; the coexistence case just doesn't seem to have come up.
- Workaround we're using meanwhile: a types-only `pnpm patch` removing the global augmentation, since we don't call `useScriptGoogleTagManager`. Happy to open a PR for whichever direction you prefer.

Togetic:
Fixes #852.

### Problem

The GTM registry augments the global `Window` with the entire `GoogleTagManagerApi`:

```ts
export interface GoogleTagManagerApi {
  google_tag_manager: GoogleTagManagerInstance
  dataLayer: DataLayer & { push: DataLayerPush }   // required
}

declare global {
  interface Window extends GoogleTagManagerApi {}
}
```

This makes `@nuxt/scripts` irreconcilable with any other package that declares `Window.dataLayer`. The common case is [`@gtm-support/core`](https://www.npmjs.com/package/@gtm-support/core) — the engine behind `@gtm-support/vue-gtm`, which a lot of Nuxt/Vue apps still use for GTM:

```ts
// @gtm-support/core/lib/index.d.ts
declare global {
  interface Window {
    dataLayer?: DataLayerObject[];   // optional, different element type
  }
}
```

The two declarations merge into one `Window`, which then fails its own `extends GoogleTagManagerApi` check. The result is `TS2430` reported at **every one of the consumer's own `Window` augmentations** — nowhere near the actual cause, and not suppressable from consumer code, since neither declaration belongs to them.

`tsc` 5.9.3 never verifies merged interfaces against their bases, so it stays silent. **TypeScript 7 (tsgo) does check, and reports it** — so this repo will hit it itself once #827 lands.

### The obvious fix does not work

#852 proposed making `dataLayer` optional. I checked, and that is **not sufficient** — there are two independent incompatibilities stacked, and making it optional only reveals the second:

```
Type 'DataLayerObject[]' is not assignable to type 'DataLayer & { push: DataLayerPush; }'.
  Type 'DataLayerObject[]' is not assignable to type '{ push: DataLayerPush; }'.
    Types of property 'push' are incompatible.
      Type '(...items: DataLayerObject[]) => number' is not assignable to type 'DataLayerPush'.
        Types of parameters 'items' and 'command' are incompatible.
          Type 'string' is not assignable to type 'DataLayerObject'.
```

`Array<T>.push` is `(...items: T[]) => number`, while `DataLayerPush`'s first overload begins `(command: string, ...)`. No amount of optionality reconciles that with a foreign element type. Any fix that keeps `dataLayer` in the global augmentation stays broken.

### Repro

Four files, `strict`, `skipLibCheck` — one file per package plus a consumer that augments `Window` (as most apps do):

<details>
<summary>files</summary>

```ts
// gtm-support.d.ts  — verbatim shape from @gtm-support/core
interface DataLayerObject extends Record<string, any> { event?: string }
declare global { interface Window { dataLayer?: DataLayerObject[] } }
export { type DataLayerObject }
```

```ts
// consumer.ts — any first-party Window augmentation
export {}
declare global { interface Window { utmData?: Record<string, string> } }
```

plus `nuxt-scripts.d.ts` carrying the registry's type block.

</details>

| Variant | TS 7.0.2 | tsc 5.9.3 |
| --- | --- | --- |
| current `main` | ❌ `TS2430` | ✅ silent (never checks) |
| `dataLayer` made optional (#852's proposal) | ❌ `TS2430` | ✅ silent |
| **this PR** | ✅ clean | ✅ clean |

### The fix

Declare only the member that is actually global and actually unambiguous:

```diff
-  interface Window extends GoogleTagManagerApi {}
+  interface Window extends Pick<GoogleTagManagerApi, 'google_tag_manager'> {}
```

### Why this is safe

- **`GoogleTagManagerApi` is untouched**, so `useScriptGoogleTagManager<T extends GoogleTagManagerApi>()` and the `use()` return type keep their exact shape. Anyone reading `dataLayer` off the returned proxy is unaffected.
- **The registry never relied on the global declaration for `dataLayer`.** It reads `(window as any)[dataLayerName]` and casts — because the name is configurable. `window.google_tag_manager` *is* read directly (`use()`), which is why that member stays declared.
- **Nothing else in this repo depends on the global `dataLayer` typing.** The four `window.dataLayer` occurrences are all inside `innerHTML` template strings (`playground/.../unhead.vue`, `scripts/generate-sizes.ts`), prose in an example (`examples/regional-consent/app.vue`), or already `(window as any)` (`test/nuxt-runtime/consent-default.nuxt.test.ts`).
- **Declaring it globally was inaccurate anyway.** The dataLayer name is configurable via the `l` / `dataLayer` options, so `window.dataLayer` is not guaranteed to exist; the typed proxy is the only access path that respects a custom name.

### Note on scope

This is a type-level change only — zero runtime bytes.

It is technically breaking for anyone who reads `window.dataLayer` directly *and* relies on this package to type it. That was arguably never a promise this package should have made (the name is configurable), and such code keeps working with a one-line local augmentation or by using the `useScriptGoogleTagManager()` proxy. Happy to put it behind a major, or to add the narrower `Window { dataLayer?: unknown }` shim instead, if you would prefer — whichever you would rather land.

I verified the above with the minimal reproduction (exact declarations from both packages, TS 7.0.2 and 5.9.3); I have not run the full repo test suite locally.

### DIFF
--- a/packages/script/src/runtime/registry/google-tag-manager.ts
+++ b/packages/script/src/runtime/registry/google-tag-manager.ts
@@ -72,9 +72,17 @@ export interface GoogleTagManagerApi {
 
 /**
  * Enhanced window type with GTM
+ *
+ * `dataLayer` is deliberately NOT declared globally. Its name is configurable through the
+ * `l` / `dataLayer` options, so `window.dataLayer` is not guaranteed to exist, and declaring it
+ * here makes this package irreconcilable with any other package that declares `Window.dataLayer`
+ * (for example `@gtm-support/core`, used by `@gtm-support/vue-gtm`): the two declarations cannot
+ * merge, so consumers of both get an unsuppressable TS2430 at every one of their own `Window`
+ * augmentations. Read it through the typed proxy returned by `useScriptGoogleTagManager()`
+ * instead, which is also the only access path that respects a custom dataLayer name.
  */
 declare global {
-  interface Window extends GoogleTagManagerApi {}
+  interface Window extends Pick<GoogleTagManagerApi, 'google_tag_manager'> {}
 }
 
 export { GoogleTagManagerOptions }
--- a/test/types/global-window.test-d.ts
+++ b/test/types/global-window.test-d.ts
@@ -1,0 +1,31 @@
+import type { GoogleTagManagerApi } from '../../packages/script/src/runtime/registry/google-tag-manager'
+import { describe, expectTypeOf, it } from 'vitest'
+
+/**
+ * Regression guard for #852.
+ *
+ * `@gtm-support/core` (the engine behind `@gtm-support/vue-gtm`) declares
+ * `Window.dataLayer?: DataLayerObject[]`. Interface declarations merge, so if this package
+ * also declares `dataLayer` on the global `Window`, the merged `Window` fails this package's
+ * own `extends` clause. TypeScript then reports TS2430 at every one of the consumer's own
+ * `Window` augmentations, which the consumer cannot suppress.
+ *
+ * `dataLayer` must stay off the global `Window`. Its name is configurable through the
+ * `l` / `dataLayer` options anyway, so `window.dataLayer` was never guaranteed to exist.
+ * The proxy returned by `useScriptGoogleTagManager()` is the supported access path.
+ */
+describe('global `Window` augmentation', () => {
+  it('does not declare `dataLayer`', () => {
+    expectTypeOf<'dataLayer' extends keyof Window ? true : false>().toEqualTypeOf<false>()
+  })
+
+  it('keeps `google_tag_manager` declared', () => {
+    // `useScriptGoogleTagManager`'s `use()` reads `window.google_tag_manager` directly.
+    expectTypeOf<Window['google_tag_manager']>().toEqualTypeOf<GoogleTagManagerApi['google_tag_manager']>()
+  })
+
+  it('still types `dataLayer` on the GTM proxy API', () => {
+    expectTypeOf<GoogleTagManagerApi['dataLayer']>().not.toBeAny()
+    expectTypeOf<GoogleTagManagerApi['dataLayer']['push']>().toBeCallableWith({ event: 'test' })
+  })
+})


harlan-zw:
### 🤖 READY · 95/100

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Human merge decision still required.

`▓▓▓▓▓ 100%`

- **Fixed:** no regression coverage. Pushed `test/types/global-window.test-d.ts`, which fails on the pre-fix declaration and passes on the fix.

Notes for the maintainer, not defects in this diff:

- The type break is real but narrow. Consumers reading `window.dataLayer` directly lose its type. Semver call is yours.
- `tsc` is not silent, contrary to the PR table. TS 5.9 reports the same TS2430 once the consumer augments `Window` from a `.ts` file. Only a `.d.ts`-only conflict is hidden by `skipLibCheck`. That makes this fix useful before TypeScript 7.
- About 25 other registry entries use the same `interface Window extends XApi {}` shape with required members, so the collision class stays open. A follow-up sweep would close it.

PR #856, opened 2026-08-08
feat: DeskCrew

webmilmind1:
Adds DeskCrew, following up on nuxt/scripts#177. Thanks for the go-ahead @harlan-zw.

DeskCrew is a support widget combining live chat, AI answers grounded in a
knowledge base, a help centre and a changelog.

## Shape

Data attributes, so it follows the Umami shape rather than a programmatic init:
`data-key` plus optional `board`, `color`, `position`, `greeting`, `launcher`.
`|| undefined` guards every optional, because with `envDefaults` an unset field
resolves to `''` and unhead renders an empty attribute onto the tag.

Synchronous `use()` rather than `resolve({ waitFor })`. I checked this against
the widget source rather than assuming: `desk.js` is a single synchronous IIFE
with the queue stub and the real API assigned at the same nesting level and no
async boundary between them, so the API exists by the time the `load` event
fires. Crisp needs `resolve()` because its API genuinely arrives after an async
ready callback; this does not.

## On the stale-reference question I raised in #177

Fixed on our side before opening this, so nothing here depends on a pending
release. The widget now assigns its methods **onto** the early stub instead of
replacing the object, so `window.DeskCrew` keeps one identity for the life of
the page and a handle taken before load stays live. It also sets a `loaded` flag
last, if a readiness signal is ever wanted.

## No bundle, no proxy, deliberately

The widget derives its API origin from the executing script's own `src`.
Bundling to `/_scripts/assets` would repoint every API call at the consuming
site's origin and break the widget; proxying would put that site's server in the
path of live chat polling. The type reflects this via
`RegistryScriptInput<..., false, false>` so the option is not even offered, and
there is a negative test in `proxy-configs` naming the reason.

## Identity is not a schema field

It is a signed per-visitor token, and everything in `nuxt.config` is a
deploy-time constant, so baking one visitor's token into a build would hand that
identity to every other visitor. It goes through `proxy.identify({ token })` at
runtime instead, documented on the docs page.

## Facade

Mirrors `ScriptCrisp`, including its slot-ordering quirk (the
`status === 'loading' || !isReady` branch making the named `#error` slot
unreachable) so the two support facades stay identical. Happy to fix both in a
follow-up, but I did not want to change Crisp's behaviour inside this PR.

Its readiness signal is the widget's Shadow DOM host (`#deskcrew-root`) rather
than script load, since load fires before anything renders and would hide the
facade over an empty screen.

## Verification, honestly

`pnpm lint` passes with zero warnings, including the `ai-deslop-*` rules, which
caught four things in my docs prose that I rewrote.

`pnpm generate:types` ran and picked the entry up.

I could **not** run `typecheck` or the test projects locally: the vitest config
loads the built `@nuxt/scripts`, and `nuxt-module-build` fails in my environment
on `loadTSCompilerOptions`. I confirmed that is not from this branch by stashing
everything and rebuilding a clean checkout, which fails identically. Relying on
CI here, and happy to fix whatever it finds.

## Not included

The `:deskcrew-demo` live component that `crisp.md` and `intercom.md` use. The
public repo's `docs/` contains only `content/`, and the app defining those demo
blocks is not here, so a maintainer would need to add it. The page is written to
read correctly without it.


harlan-zw:
### 🤖 WAITING · maintainer decision on registry inclusion

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Human merge decision still required. Waiting: the cited go-ahead could not be found.

`▓▓▓▓▓ 100%`

- **Open:** The description cites a go-ahead in nuxt/scripts#177. That discussion contains no DeskCrew request and no maintainer reply about DeskCrew, and the author has no earlier activity in this repository. Adding a registry entry blesses a new third-party origin, so this premise needs confirming before merge. Next: @harlan-zw to accept or decline DeskCrew for the registry.

No code defects found. The adversarial pass drove the live `deskcrew.io/desk.js` in a browser and confirmed the two claims the integration rests on: `window.DeskCrew` is fully wired with the exact documented method set by the `load` event, so synchronous `use()` is correct, and `#deskcrew-root` exists with its shadow root and launcher, so the facade readiness signal holds. Schema, capability flags and the negative proxy test match how the widget resolves its API origin.

PR #872, opened 2026-08-20
fix(carbon-ads): keep the script id resolvable after unmount

harlan-zw:
### 📚 Description

Unmounting `ScriptCarbonAds` crashes the page. carbon.js resolves `#_carbonads_js` from its own pending ad callback and reads `.src` off it with no null check, so removing that element throws inside third-party code where nothing can catch it. Sites that key the ad on the route to refresh it per page hit this on every navigation.

The root ref was also seeded from `document.getElementById('_carbonads_js')`, which is the script's id rather than a container. Every other registry component starts that ref at `null`.

Open question: two instances mounted at once still collide on the hardcoded id. Carbon's own code assumes one per page, so I left it alone.

### DIFF
--- a/packages/script/src/runtime/components/ScriptCarbonAds.vue
+++ b/packages/script/src/runtime/components/ScriptCarbonAds.vue
@@ -26,7 +26,7 @@ defineSlots<{
 }>()
 
 const attrId = `_carbonads_js`
-const carbonadsEl = ref<HTMLElement | null>(import.meta.client ? document.getElementById(attrId) : null)
+const carbonadsEl = ref<HTMLElement | null>(null)
 // syncs to useScript status
 const status = ref('awaitingLoad')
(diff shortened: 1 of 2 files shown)


harlan-zw:
Checked in a real browser against live carbon.js, since CI has no Carbon coverage.

Mount, unmount, remount, then three `:key` cycles on a page that mirrors how unlighthouse.dev uses the component:

Without this change, after unmount:
  0 elements match #_carbonads_js
  window._carbonads.getUrlVar('serve') -> TypeError: Cannot read properties of null (reading 'src')

With it, after unmount:
  1 element matches #_carbonads_js
  window._carbonads.getUrlVar('serve') -> 'CW7DTKJL'

Remount and the three key cycles kept exactly one script element and one rendered ad, with no page errors.

PR #876, opened 2026-08-21
fix(proxy): disable proxying for static deployments

### LINKED ISSUE #875: Cloudflare Web Analytics in SSR:false
### 📚 What are you trying to do?

Set Cloudflare Web Analytics in SSR:false nuxt.

### 🔍 What have you tried?

I followed https://scripts.nuxt.com/scripts/cloudflare-web-analytics and added to my nuxt.config:

```
  scripts: {
    registry: {
      // NUXT_PUBLIC_SCRIPTS_CLOUDFLARE_WEB_ANALYTICS_TOKEN
      cloudflareWebAnalytics: { trigger: 'onNuxtReady', spa: true },
    }
  },
```
And I added `NUXT_PUBLIC_SCRIPTS_CLOUDFLARE_WEB_ANALYTICS_TOKEN` as an env var.

I am using `ssr:false` and with `nuxi generate` to deploy in cloudflare workers. But when deploying it, it's trying to hit  `<domain>/_scripts/p/cloudflareinsights.com/cdn-cgi/rum` which is giving 405, and I believe is wrong.

Shouldn't nuxt's `ssr:false` (and/or script's `spa:true`) make it just attach the script like the commented sample in https://github.com/nuxt/scripts/blob/main/packages/script/src/runtime/registry/cloudflare-web-analytics.ts since has no server or route?

In the meantime, I'm adding it directly and is working:

```
  app: {
    head: {
      script: [
        {
          src: 'https://static.cloudflareinsights.com/beacon.min.js',
          'data-cf-beacon': '{"token": "MYTOKEN"}',
          defer: true,
        },
      ],
```

Thanks for any help.

harlan-zw:
### 🔗 Linked issue

Resolves #875

### 📚 Description

A `nuxi generate` site has no Nitro server to answer `/_scripts/p/**`. The module still rewrote bundled scripts to those URLs, so beacons hit `<domain>/_scripts/p/cloudflareinsights.com/cdn-cgi/rum` and static hosting answered 405. The existing static warning only read `NITRO_PRESET`, which `nuxi generate` never sets, so the breakage shipped silent.

Static output is now detected from `_generate`, `nitro.static`, and the static Nitro presets. Proxy rewrites, the intercept plugin, auto-injected proxy endpoints, and the proxy route are skipped, and one warning lists the affected scripts. Scripts still bundle and load from your own domain. Collection requests keep their original third-party URLs and go direct, which is what Cloudflare Web Analytics needs on static hosting.

### DIFF
--- a/docs/content/docs/1.guides/2.first-party.md
+++ b/docs/content/docs/1.guides/2.first-party.md
@@ -236,23 +236,11 @@ PostHog receives the proxy endpoint through SDK config, so it can proxy collecti
 
 ### Static Hosting (SSG)
 
-The reverse proxy requires a **server runtime**. A fully static deployment serves the output of [`nuxt generate`](https://nuxt.com/docs/getting-started/prerendering) without a Nitro process to handle `/_scripts/p/**`. Nuxt Scripts warns for known static presets, but it does not rewrite proxy URLs to their third-party origins. Disable proxying for affected scripts or use a host that supports external-origin rewrites. For example, [Vercel rewrites](https://vercel.com/docs/routing/rewrites) accept `/:path*` captures and external destinations:
+The reverse proxy requires a **server runtime**. A fully static deployment serves the output of [`nuxt generate`](https://nuxt.com/docs/getting-started/prerendering) without a Nitro process to handle `/_scripts/p/**`.
 
-```json [vercel.json]
-{
-  "rewrites": [
-    { "source": "/_scripts/p/www.google-analytics.com/:path*", "destination": "https://www.google-analytics.com/:path*" },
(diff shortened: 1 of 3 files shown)


harlan-zw:
Checked by hand on the playground, since CI does not inspect generated assets:
- `nuxi generate`: the bundled beacon (`_scripts/assets/e1ac4849b9d4a498.js`) has zero `/_scripts/p` references and keeps the direct `cdn-cgi/rum` endpoint, so the 405 from #875 is gone
- `nuxi generate` prints the new warning listing every affected script
- `nuxi build` (server runtime): proxy config still lands in the Nitro server bundle and the beacon is still rewritten to `/_scripts/p`, so first-party mode is unchanged

PR #878, opened 2026-08-21
fix(registry): don't declare the global `Window` via `extends`

Togetic:
Follow-up sweep for #852, invited by the triage note on #855:

> About 25 other registry entries use the same `interface Window extends XApi {}` shape with required members, so the collision class stays open. A follow-up sweep would close it.

#855 fixed the GTM entry by narrowing *which* members reach the global `Window`. This closes the class by changing the *shape* that makes a collision catastrophic in the first place.

### The defect class

`interface Window extends XApi {}` is the part that turns a routine member collision into an unsuppressable failure. Interface declarations merge, so when any other package declares one of the same members, the merged `Window` stops satisfying the `extends` clause this package added — and TypeScript reports `TS2430` at **every** `Window` augmentation in the program, including the consumer's own. Those are nowhere near the cause, and `skipLibCheck` cannot silence them because they are the consumer's own `.ts` files.

Declaring the same members inline removes the clause that can fail:

```diff
-  interface Window extends PayPalApi {}
+  interface Window {
+    paypal: PayPalApi['paypal']
+  }
```

A genuine collision then surfaces as `TS2687`/`TS2717` **on the two conflicting declarations**, which are both in `.d.ts` files and therefore fall under `skipLibCheck` like any other dependency-vs-dependency disagreement — diagnosable, and pointing at the actual cause.

### This is not hypothetical — there is a second live instance

`@paypal/paypal-js` (the official PayPal SDK types, and a direct devDependency of this repo) declares:

```ts
// @paypal/paypal-js@10.1.0 types/index.d.ts
declare global {
  interface Window { paypal?: PayPalNamespace | null }
}
```

The registry declared `paypal: PayPalV6Namespace` — required, and a different namespace type. Same two stacked incompatibilities as `@gtm-support/core`. Verified against the real package, `strict`, `skipLibCheck: true`, with a consumer that augments `Window` from a `.ts` file:

| | tsgo 7.0.0-dev | tsc 6.0.3 |
| --- | --- | --- |
| before | ❌ `TS2430` at `consumer.ts` | ✅ clean |
| after | ✅ clean | ✅ clean |

### Correction to the framing in #855

The triage note on #855 said:

> `tsc` is not silent, contrary to the PR table. TS 5.9 reports the same `TS2430` once the consumer augments `Window` from a `.ts` file. Only a `.d.ts`-only conflict is hidden by `skipLibCheck`.

I could not reproduce that, and measured the opposite. On **TS 5.9.3 and 6.0.3** the `TS2430` anchors to `lib.dom.d.ts` — the primary `Window` declaration — so `skipLibCheck: true` suppresses it **regardless** of whether the consumer augments `Window` from a `.ts` file, a `.d.ts` file, or with its own `extends` clause. It surfaces pre-TS7 only with `skipLibCheck: false`. On **tsgo** the error is instead attributed to every augmentation site, including the consumer's own `.ts` files, which is why `skipLibCheck` stops helping. That matches the original report in #852.

The conclusion still holds, but for a different reason, and it is worth stating precisely because it bounds what this PR fixes:

- **Pre-TS7**, the foreign *optional* declaration silently wins the merge, so consumers reading `window.paypal` get `TS2722`/`TS18048` "possibly undefined" at each read site. **This PR does not change that** — it is inherent to declaration merging, and it is equally true before and after.
- **On TS7**, the collision additionally becomes an unsuppressable `TS2430` at every consumer `Window` augmentation. **That is what this PR eliminates.**

Happy to be shown the configuration behind the original note if I've missed one — the repro is four files and I can push it as a fixture.

### Scope

20 entries, all in `packages/script/src/runtime/registry/`: `ahrefs-analytics`, `calendly`, `clarity`, `cloudflare-web-analytics`, `google-adsense`, `google-recaptcha`, `hotjar`, `intercom`, `leaflet`, `linkedin-insight`, `matomo-analytics`, `meta-pixel`, `paypal`, `reddit-pixel`, `segment`, `snapchat-pixel`, `vimeo-player`, `x-pixel`, `youtube-player` — plus `google-tag-manager`, because #855's `Pick<GoogleTagManagerApi, 'google_tag_manager'>` narrowed the surface but kept the `extends` clause, so that entry still carried the shape for its remaining member.

The exact figure is 19 unfixed entries rather than ~25. The other `Window` augmentations in the registry (`bing-uet`, `crisp`, `databuddy-analytics`, `deskcrew`, `fathom-analytics`, `lemon-squeezy`, `maplibre`, `mixpanel-analytics`, `plausible-analytics`, `posthog`, `rybbit-analytics`, `speedcurve`, `tiktok-pixel`, `umami-analytics`, `usercentrics`, `vercel-analytics`) already declare members inline, so they cannot produce `TS2430` and are left untouched.

### Semver — your call

**As written, this is not a type break.** Every member keeps its exact type and its exact required/optional modifier; `Window['fbq']` and friends resolve to what they resolved to before. That is the reason I did not port #855's `Pick<>` literally: dropping members across 19 entries *would* be the broad break the triage note anticipated, and unlike GTM's `dataLayer` — reached through `(window as any)[dataLayerName]` because its name is configurable — most of these members are read off bare `window` by the registries themselves, so removing them would also break this package's own build.

There is a stronger version available if you want it: for entries whose member is genuinely not guaranteed to exist under that name, drop it from the global `Window` GTM-style and let the `useScript*()` proxy be the only typed access path. That *is* a break for anyone reading `window.<prop>` directly, and it is much broader than it was for GTM alone. I have deliberately not made that call — say the word and I will do it as a follow-up, behind whatever version you want.

Worth noting either way: `segment` puts six very generic names on the global `Window` — `track`, `page`, `identify`, `group`, `alias`, `reset`. Those are the most likely of the whole registry to collide with something. This PR does not change that, but it does mean a collision on them no longer takes the consumer's build down at a distance.

### Tests

Extends `test/types/global-window.test-d.ts` (the file the bot pushed onto #855) rather than adding a harness:

- every swept member is pinned to its API interface with `toEqualTypeOf`, so the rewrite stays type-identical to the `extends` form it replaced — this is the guard on the "not a type break" claim above;
- a drift guard asserts `Exclude<keyof XApi, keyof Window>` is `never` for each entry, since inline lists no longer track the API automatically the way `extends` did. `GoogleTagManagerApi` is exempted, as `dataLayer` is deliberately not global per #855. Verified to fail when a member is removed from a `Window` declaration.

Type-level assertions cannot fail on the pre-fix code here — that is precisely what "type-identical" means — so the invariant itself is guarded by `test/unit/global-window-augmentation.test.ts`, which scans the registry sources and asserts none declares `Window` via an `extends` clause. It fails on `main` listing all 20 offenders, and passes on this branch.

### Verification

`pnpm lint`, `pnpm typecheck`, and `pnpm vitest run --project typecheck --project unit` (78 files, 941 tests) all pass. Not run locally: `e2e` (needs browsers) and `build`.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/script/src/runtime/registry/ahrefs-analytics.ts
+++ b/packages/script/src/runtime/registry/ahrefs-analytics.ts
@@ -29,7 +29,11 @@ export interface AhrefsAnalyticsApi {
 }
 
 declare global {
-  interface Window extends AhrefsAnalyticsApi {}
+  // Declared inline rather than via `extends`: an `extends` clause on the global `Window`
+  // surfaces as an unsuppressable TS2430 in consumer code when another package declares it (#852).
+  interface Window {
+    AhrefsAnalytics: AhrefsAnalyticsApi['AhrefsAnalytics']
+  }
 }
 
 /**
--- a/packages/script/src/runtime/registry/calendly.ts
+++ b/packages/script/src/runtime/registry/calendly.ts
@@ -76,7 +76,11 @@ export interface CalendlyApi {
 }
 
 declare global {
-  interface Window extends CalendlyApi {}
+  // Declared inline rather than via `extends`: an `extends` clause on the global `Window`
+  // surfaces as an unsuppressable TS2430 in consumer code when another package declares it (#852).
+  interface Window {
+    Calendly: CalendlyApi['Calendly']
+  }
 }
 
 const CALENDLY_CSS_KEY = 'nuxt-scripts-calendly-css'
--- a/packages/script/src/runtime/registry/clarity.ts
+++ b/packages/script/src/runtime/registry/clarity.ts
@@ -25,7 +25,11 @@ export interface ClarityApi {
 }
 
 declare global {
-  interface Window extends ClarityApi {}
+  // Declared inline rather than via `extends`: an `extends` clause on the global `Window`
+  // surfaces as an unsuppressable TS2430 in consumer code when another package declares it (#852).
+  interface Window {
+    clarity: ClarityApi['clarity']
+  }
 }
 
 export type ClarityInput = RegistryScriptInput<typeof ClarityOptions>
--- a/packages/script/src/runtime/registry/cloudflare-web-analytics.ts
+++ b/packages/script/src/runtime/registry/cloudflare-web-analytics.ts
@@ -20,7 +20,11 @@ export interface CloudflareWebAnalyticsApi {
 }
 
 declare global {
-  interface Window extends CloudflareWebAnalyticsApi {}
+  // Declared inline rather than via `extends`: an `extends` clause on the global `Window`
+  // surfaces as an unsuppressable TS2430 in consumer code when another package declares it (#852).
+  interface Window {
+    __cfBeacon: CloudflareWebAnalyticsApi['__cfBeacon']
+  }
 }
 
 export type CloudflareWebAnalyticsInput = RegistryScriptInput<typeof CloudflareWebAnalyticsOptions>
--- a/packages/script/src/runtime/registry/google-adsense.ts
+++ b/packages/script/src/runtime/registry/google-adsense.ts
@@ -15,7 +15,11 @@ export interface GoogleAdsenseApi {
 }
 
 declare global {
-  interface Window extends GoogleAdsenseApi {}
+  // Declared inline rather than via `extends`: an `extends` clause on the global `Window`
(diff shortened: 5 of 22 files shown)


harlan-zw:
Awesome, thanks for that :)

PR #886, opened 2026-08-25
feat: Tawk.to

atlaxt:
### 📚 Description

[Tawk.to](https://www.tawk.to/) is a free live chat widget. @danielroe and @harlan-zw suggested pulling it into the registry from my `nuxt-tawk-to` module, so this adds `tawkTo` / `useScriptTawkTo()`.

Two things about Tawk shape the API. Its embed script dispatches `window` CustomEvents (`tawkLoad`, `tawkStatusChange`, `tawkChatMaximized`, and so on) alongside the documented `Tawk_API.onXxx = fn` callbacks, so the composable bridges those into five readonly refs and twenty typed listeners instead of leaving you on `addEventListener`. Separately, unhead's script proxy is fire-and-forget: it queues and replays calls but discards the return value, and it has no `set` trap. Real getters like `getStatus()` and the `visitor` setter therefore live on the composable's return value, calling straight through to `window.Tawk_API`, rather than on `proxy`.

Loose ends:

- Bundle and proxy are both off. I have not verified how the embed resolves its own API origin, or what live-chat polling connections a proxy would end up in front of, so the capabilities are undeclared rather than guessed.
- `setVisitor()` only works before the embed script is requested, because Tawk ignores `Tawk_API.visitor` after that. Past that point the composable warns and drops the write, and the docs point at `setAttributes()` instead.
- The registry entry points at `tawk.to/icon.png` rather than inlining a mark. Tawk publishes no SVG anywhere I could find: not on the site, not in the widget bundle, not in the dashboard CSS, and not in any of their six official platform plugin repos. `clarity` and `pulseAnalytics` already take an image URL for the same reason.


harlan-zw:
### 🤖 REVIEW PAUSED

This pull request is from an outside contributor. Add the `harlan-agent-review` label to approve automated review and verified repairs for head commit `7d8be5125066`.

PR #893, opened 2026-09-08
fix(proxy): follow allowlisted upstream redirects

### LINKED ISSUE #885: 502 Unsafe upstream redirect when Meta Pixel's /tr endpoint returns 302
### 🐛 The bug

**Env**
 `@nuxt/scripts@1.3.5`
 `nuxt@4.4.8`

**Config**: 
```
scripts: {
    registry: {
      googleTagManager: {
        id: GTM_ID,
        trigger: 'onNuxtReady',
        bundle: false,
      },
      metaPixel: {
        id: META_PIXEL_KEY,
        trigger: 'onNuxtReady',
      },
      googleAnalytics: {
        id: GOOGLE_ANALYTICS_KEY,
        trigger: 'onNuxtReady',
      },
    },
  },
```
Intermittently, POST `/_scripts/p/www.facebook.com/tr/` fails with a **502** Unsafe upstream redirect because Facebook responds with a **302** for this particular request, and the first-party proxy handler refuses to follow it.

In proxy-handler.ts, the fetch to the upstream is always made with redirect: `manual`, and any **3xx** response other than **304** throws a **502**:

```
const requestInit: RequestInit = {
  ...
  redirect: 'manual',
}
...
if (response.status >= 300 && response.status < 400 && response.status !== 304) {
  throw createError({
    statusCode: 502,
    statusMessage: 'Unsafe upstream redirect',
    message: 'Proxy upstream returned a redirect that was not followed',
  })
}
```
I understand this is intentional — following the Location header would mean sending a request to a host that was never validated against the domain allowlist (**SSRF** risk), so blindly following it isn't safe. I noticed image-proxy.ts already has a followRedirects option that re-validates each hop against the same allowlist before following it — is something similar planned/possible for the event-collection proxy handler used by registry scripts like Meta Pixel?

Image

Image

### 🛠️ To reproduce

-

### 🌈 Expected behavior

-

### ℹ️ Additional context

_No response_

harlan-zw:
### 🔗 Linked issue

Closes #885

### 📚 Description

Meta Pixel events were silently dropped: the proxy answered Facebook's intermittent 302 from `/tr` with a 502 Unsafe upstream redirect, so the beacon never reached its destination. This starts from issue #885.

Redirects are now followed the way a browser fetch would follow them, but only after each hop passes the same HTTPS, public network, and domain allowlist checks as the original target, with a limit of five hops. A 302 replays as GET without the body and a 307 keeps it, matching the fetch spec.

One tradeoff to flag: proxied request bodies are now buffered under the same 2 MiB cap the privacy transforms already enforced, so an oversized beacon gets a 413 instead of streaming through.

### DIFF
--- a/packages/script/src/runtime/server/proxy-handler.ts
+++ b/packages/script/src/runtime/server/proxy-handler.ts
@@ -30,6 +30,7 @@ const COMPRESSION_RE = /gzip|deflate|br|compress|base64/i
 const CLIENT_HINT_VERSION_RE = /;v="(\d+)\.[^"]*"/g
 const MAX_TRANSFORM_BODY_SIZE = 2 * 1024 * 1024
 const UPSTREAM_TIMEOUT_MS = 15000
+const MAX_UPSTREAM_REDIRECTS = 5
 const SKIP_RESPONSE_HEADERS = new Set([
   'alt-svc',
   'clear-site-data',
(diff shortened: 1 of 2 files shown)


harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-09 03:29 UTC.

GitHub merged this pull request. No further automated Review will run.

PR #901, opened 2026-09-16
fix(maplibre): surface layer errors and stop needless source rebuilds

harlan-zw:
### 📚 Description

Two things I hit while building a clustered map on the MapLibre components.

A typo in a paint expression gives you a basemap with no data on it. `useMapLibreResource` caught the failure, ran `console.error`, and stopped: no emit, no rethrow, no state change. The rethrow also left the resource unset, so fixing the expression never rebuilt the layer, the component was just dead until a remount. `ScriptMapLibreGeoJson` emits `error` now, and a failed first sync keeps its style and prop listeners so a correction recovers on its own. The `console.error` stays; being *only* a `console.error` was the problem.

The second one is the footgun. The rebuild watcher compared `layers` and `sourceOptions` by identity, and an inline `:layers="[...]"` array literal gets a fresh identity on every parent render. So any unrelated reactive change in the parent called `removeLayer` for each layer, then `removeSource`, then added everything back. On a clustered source that throws away the cluster index and visibly flashes the map. The cheapest-looking usage was the one that triggered it. It compares content now.

I stopped short of the rest of the fix. Routing cluster changes to `source.setClusterOptions()` and diffing layers against `ownedLayerIds` instead of dropping the source would be better, but layer diffing has to reason about ordering, `beforeId`, type changes and paint versus layout versus filter, and `setClusterOptions()` is async against a synchronous rebuild path. The content comparison alone kills the reported symptom. Happy to take the rest as a follow-up.

Non-breaking. `error` is a new emit, `onError` is a new optional option, and every rebuild that used to happen on a real content change still happens.


harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-16 06:13 UTC.

e9b38c5c-4a1f-4e32-b008-7cfede36dbdc

PR #902, opened 2026-09-16
docs(maplibre): document clustering and raw map instance usage

harlan-zw:
### 📚 Description

I was building a clustered point map on the MapLibre components, and wanted to know whether `ScriptMapLibreGeoJson` clusters or whether I had to add the source myself. The docs cannot answer that. Across all ten MapLibre pages, `cluster`, `setFilter`, `setFeatureState`, `queryRenderedFeatures`, `fitBounds`, `easeTo`, `flyTo`, `style.load`, `feature-state` and `hover` appear zero times. The GeoJSON API page was 14 lines and said `sourceOptions` is "GeoJSON source options. `type` and `data` are supplied by the component."

It does cluster. `sourceOptions` is `Omit<GeoJSONSourceSpecification, 'type' | 'data'>`, so it carries `cluster`, `clusterRadius`, `clusterMaxZoom`, `clusterMinPoints`, `clusterProperties`, `promoteId`, `generateId` and `filter`. I only found that by resolving the `Omit` through `maplibre-gl` into `@maplibre/maplibre-gl-style-spec`. The most useful thing the component does is invisible.

Then I counted what a real data-heavy map needs. Seven things. The components cover one of them outright, the clustered source itself. The other six land on the raw `maplibregl.Map`: expanding a cluster on click, reactive filtering, hover highlight, a popup from a layer click, fitting the camera to a filtered set, and reading the features in view. That split is a fine design. Nothing in the docs says it, so you work it out by reading `ScriptMapLibreMap.vue`.

So, two additions.

The **GeoJSON API page** gets a table of every `sourceOptions` field, and a clustering example: `cluster: true` with `clusterRadius` and `clusterProperties`, the three-layer `point_count` pattern, and a `@ready` handler that reads `getClusterExpansionZoom` off the source and eases into the cluster.

Two things in that example are load-bearing and both got a callout:

- `layers` and `source-options` are declared as constants, not written inline. An inline literal is a new identity on every parent render, `syncResources` sees the change, and the source is removed and re-added. On a clustered source that throws away the cluster index and flashes the map. Docs examples are what people copy, so showing the inline form would hand everyone the footgun.
- `text-font` names `Noto Sans Bold`, which OpenFreeMap actually serves. Name one it does not serve and v6 draws the text with a local browser font through TinySDF. The label still renders, in the wrong typeface, with a single `warnOnce` in the console.

Then a **new Raw Map Instance guide**, because six of seven is worth a page of its own. It covers getting the `Map` out of `ready`, that `ready` fires inside `map.once('load', ...)` so `addSource` is safe there, and that `setStyle` drops everything you added yourself so it needs re-adding on `style.load`. After that, `setFilter`, `setFeatureState` with `promoteId`, `queryRenderedFeatures`, `fitBounds` and `easeTo`.

One trap in there caught me in the v6 source. `loadMatchingFeature` guards with `if (id && sourceFeatureState)`, so a feature id of `0` or `''` comes back from `queryRenderedFeatures` with empty state. The GPU paint path has no such guard, so the map still draws the hover correctly and only the handler that reads `event.features[0].state` breaks. Took me a second pass through the bundle to separate those two, and the callout says exactly that much and no more.

Loose ends I did not take:

- The example annotates its layers as `Omit<CircleLayerSpecification, 'source'> | Omit<SymbolLayerSpecification, 'source'>`, because `ScriptMapLibreGeoJsonLayer` is not importable from any public entry point. Without an annotation the expression arrays widen to `string[]` and the sample does not compile. Fix the export and the example gets shorter.
- `ScriptMapLibreGeoJsonLayer` is `Omit<LayerSpecification, 'source'>`, and non-distributive `Omit` over that union drops `filter`, since background layers have no filter. The prop works only because it is passed as a variable, so excess property checking never fires. Uncomfortable to build a documented example on, but the type is a source change and out of scope here.
- An empty-string feature id hashes to numeric `0` in `FeaturePositionMap`, so the two collide on the paint path as well. Obscure enough that I left it out of the callout. Say the word if you want it in.
- Docs only, so the `CI` workflow skips this under its `paths-ignore: docs/**`. CodeQL and the nightly build still run. Review is the real gate.


harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-16 06:31 UTC.

The pull request closed.

PR #903, opened 2026-09-16
feat(maplibre): add layer events, camera methods and style lifecycle

harlan-zw:
Follows #901, which is merged. Base retargeted to `main`.

### 📚 Description

Same clustered map as #901. Of the seven things that map needed, six ended up on the raw `maplibregl.Map`. These four are the ones that should never have gone there.

**Layer events.** `ScriptMapLibreGeoJson` had no `defineEmits` at all, while `ScriptMapLibreMarker` gets first-class `click` and drag events. Layers are how you render any real dataset, so cluster click, point click and hover all meant `map.on('click', layerId, handler)` on the raw instance plus your own `Subscription` bookkeeping. It emits `click`, `mouseenter` and `mouseleave` for the layers it owns now. There is also a `cursor` prop, which is the other line of boilerplate everyone writes.

**Camera.** All four camera watchers hard-coded `jumpTo` and the expose was `{ maplibre, map, load }`, so `v-model:center` and `v-model:zoom` could only teleport and fitting to a bbox had no declarative form. `fitBounds`, `easeTo` and `flyTo` are on the expose, so they compose with the `ready` payload people already destructure. The props still use `jumpTo`.

**Style lifecycle.** The map re-emitted nine events but not `load` or `style.load`, and a `mapStyle` change called `setStyle()` with no signal. MapLibre drops every source and layer on `setStyle`, so a light and dark basemap toggle silently loses anything you added on the raw map. It emits `load` and `styleload` now.

**Types.** The prop, emit, expose and layer types were declared in the SFCs with no public route. `exports` lists only `.`, `./registry`, `./stats` and `./types-source`, so the only way in was `#nuxt-scripts/components/MapLibre/ScriptMapLibreGeoJson.vue`, which reads private and ends in `.vue`. There was no route at all to `ScriptMapLibreGeoJsonLayer`, which is the one you actually need. They live in `runtime/components/MapLibre/types.ts` now, re-exported from both `@nuxt/scripts` and `#nuxt-scripts/types`, following the `components/GoogleMaps/types.ts` pattern already in the tree.

Two follow-on changes that came with the move:

- `scripts/generate-registry-types.ts` reads a vendor's sibling `types.ts`. Without it the docs tables for these components lost 325 lines the moment the props left the SFC.
- `addComponentsDir` is limited to `.vue`. `GoogleMaps/types.ts` was already registering as a global `Types` component, and a second `types.ts` turns that into an error.

Two commits at the end fix `ScriptMapLibreGeoJsonLayer` itself.

The first: it was `Omit<LayerSpecification, 'source'>`, and `Omit` is not distributive, so it collapsed the union to its common keys. `BackgroundLayerSpecification` has no `filter`, so `filter` was not in the type at all, and paint and layout expressions lost per-layer narrowing. Nobody noticed because you pass the array as a variable and excess property checking never fires. Write `filter: ['has', 'point_count']` inline, which is every clustering example ever written, and it does not compile. MapLibre exports `DistributiveOmit` for exactly this.

The second: restoring the union exposed that the type accepted a `raster`, `hillshade`, `color-relief` or `background` layer with no `source`, and the component then handed it the GeoJSON source id. MapLibre's own style validation rejects all of those. `hillshade` and `color-relief` need a `raster-dem` source, `raster` needs raster tiles, and `background` takes no source at all. One of them threw and the rollback dropped every layer the component owned. Source-less layers are limited to the six kinds a GeoJSON source can render.

### ⚠️ Breaking Changes

Two narrow type breaks, both on `ScriptMapLibreGeoJsonLayer`. Neither changes runtime behaviour, and both reject shapes that could not work.

1. An untyped hoisted layer array whose `filter` infers as `(string | string[])[]` no longer assigns.
2. A layer array holding a source-less `raster`, `hillshade`, `color-relief` or `background` layer no longer assigns.

### 📝 Migration

For the first, annotate the array or add `satisfies`:

```ts
const layers: ScriptMapLibreGeoJsonLayer[] = [/* … */]
```

`playground/pages/third-parties/maplibre.vue` was the only place in this repo that needed it, and it is annotated here.

For the second, give the layer its own source, or move it out of `<ScriptMapLibreGeoJson>` and add it on the raw map. Nothing in this repo needed it.


harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-16 08:12 UTC.

2cfa7152-6bc9-46ac-825d-3e17790cb521

PR #905, opened 2026-09-16
bug report: pre-load analytics events are sent twice (do not merge as-is)

harlan-zw:
### 📚 Description

**Do not merge. This branch is now empty; it is a written-up bug report, not a fix.**

Pulse and Rybbit each send a `track()` / `event()` call made before the script loads **twice**. Measured in a real browser against the live vendor endpoints, one pre-load call:

- Pulse: 2 POSTs to `pulse-api.ciphera.net/api/v1/events`
- Rybbit: 2 POSTs to `app.rybbit.io/api/track`

Both entries buffer pre-load calls on `globalThis` behind a `Symbol` and flush them from `use()`. My first attempt deleted those buffers, on the theory that the script proxy already replays. That is wrong, and browser testing caught it: with the buffers gone, **three** distinct pre-load calls delivered **one** event. The buffers are load-bearing.

The real cause is in `replayRecordings` in unhead:

```js
} else if (type === "apply") {
  context = context.call(prevContext, ...args)
}
```

`context` is reassigned to the call's return value. For a void method like `track()` that is `undefined`, so a second `apply` in the same recording has nothing to call. `createScriptProxy` caches child nodes in a `Map`, so repeated `proxy.track(...)` calls land in one recording rather than one each. The replay therefore delivers the first call and stops.

That yields the counts observed with the current code, three pre-load calls:

| source | delivered |
|---|---|
| the entry's own buffer | 3 |
| the proxy's replay | 1 |
| **total** | **4** |

So the duplicate is the proxy's single extra replay landing on top of a buffer that already did the job.

Fixing it in the registry entries means suppressing that one replayed call, which is guesswork about which call it was. The clean fix is upstream, either replaying each `apply` against the original `prevContext` instead of the previous return value, or not collapsing repeated calls into one recording. Worth confirming against unhead before anyone writes registry-level workarounds.

Reproduction: a page that calls `proxy.track()` three times during `setup()` with `trigger: 'manual'`, then loads the script and counts outbound requests.

> 🤖 AI disclosure: [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) wrote this description. [My AI open-source policy](https://harlanzw.com/blog/ai-in-open-source).


harlan-zw:
Retracting this. There is no duplication bug. The premise was a measurement error on my part.

Both vendors POST their automatic pageview to the **same endpoint** as custom events, so I counted it as a duplicated `track()`. Capturing the request bodies shows what the two POSTs actually are, for a single pre-load `track('p_only')`:

```
[0] name=undefined  keys=domain,url,title,referrer,screen,language,timezone,hs,client_os_hint   <- pageview
[1] name="p_only"   keys=domain,url,title,referrer,screen,name                                  <- the event, once
```

Three distinct pre-load calls give `[pageview, p_one, p_two, p_three]`. Rybbit behaves the same way: three calls produce four `/api/track` POSTs, which is N+1 for a constant session request, not 2N for duplication.

The existing `Symbol`-keyed queues in both entries are correct and load-bearing. `proxy.track` resolves to the composable's own function rather than being recorded by the script proxy, so that queue is the only thing capturing calls made before load. The change I originally proposed here would have silently dropped every pre-load analytics event, which browser testing caught before it went anywhere.

For the record, unhead's proxy replay is also fine. `createScriptProxy` records one stack entry per root property access and `replayRecordings` replays all of them in order, verified directly:

```
recordings: [[get track, apply["one"]], [get track, apply["two"]], [get track, apply["three"]]]
delivered:  3  [["one"],["two"],["three"]]
```

No change needed in this repo or upstream. Closing.

> 🤖 [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this. [My AI open-source policy](https://harlanzw.com/blog/ai-in-open-source).

PR #906, opened 2026-09-16
fix(pulse-analytics): keep draining the queue when a replayed call throws

uz1mani:
Follow-up to #899, for the note CodeRabbit left after the merge.

`use()` sets `flushed` and then replays the queued `track()` calls with `shift()`. If the tracker threw synchronously on one of them, the loop exited and every event queued after it was stranded for good: `track()` dispatches directly once the tracker is ready, so nothing would drain the queue again. Each replay is now wrapped so one rejected event cannot take the rest with it. Direct dispatch is unchanged.

Test added to `test/unit/pulse-analytics-queue.test.ts`: first replayed call throws, the other two still land, the queue ends empty, `use()` does not throw. Verified red against the previous loop, green with the fix.


harlan-zw:
### 🤖 REVIEW PAUSED

This pull request is from an outside contributor. Add the `harlan-agent-review` label to approve automated review and verified repairs for head commit `e537c33cd23f`.

uz1mani:
Unrelated to this fix, but while you're here: any idea when the next beta goes out? Pulse is on scripts.nuxt.com already, though only the page itself. The card, logo and stats come from the published package, so right now it sits under Utility with nothing on it. No rush, I just want to know whether to point our own docs at it now or hold off a bit.

harlan-zw:
2.0.0-beta.7 is out now :) Not sure if docs will properly reflect it's v2 only atm

uz1mani:
Perfect, thanks. Confirmed beta.7 carries it (the registry file is in the published tarball). The site still files Pulse under Utility without the card for now, so I'd guess it just needs a redeploy against beta.7 whenever that next happens. Nothing needed from me, and thanks for the quick turnaround on both.

PR #907, opened 2026-09-16
fix(maplibre): re-add the GeoJSON source after a style swap and diff layers

harlan-zw:
Stacked on #903 (`maplibre-layer-events-camera-style`). Review that first; this branch adds only the two commits above its head.

Both papercuts were found while dogfooding, by building a real clustered map app on `@nuxt/scripts@2.0.0-beta.6`. Neither came from an audit.

Both changes are **non-breaking**. No prop, emit or expose is removed or renamed.

## PC-9: the GeoJSON source was lost for good on a style swap

**Symptom.** A light and dark basemap toggle made every point disappear and never come back. The map showed a basemap with no data and no error.

`ScriptMapLibreGeoJson` already registered `map.on('style.load')` to re-add its source and layers. `syncResources` then returned early on `!map.isStyleLoaded()`, and nothing retried.

The readiness test was wrong. MapLibre accepts `addSource` and `addLayer` from `style.load` onwards, because `Style._checkLoaded()` only requires `Style._loaded`. `map.isStyleLoaded()` maps to `Style.loaded()`, which additionally requires no pending source updates, every tile manager loaded, and the sprite loaded. It is still false when `style.load` fires.

Live probe from the browser, registered before the swap on MapLibre 6.10.0:

```
styledata isStyleLoaded=false layer=true
style.load isStyleLoaded=false
style.load isStyleLoaded=false
styledata isStyleLoaded=false layer=false
idle      isStyleLoaded=true  source=false
```

By `idle` the style really is loaded and the source is gone.

**Ordering confirmed against the installed MapLibre v6 source**, not the probe alone:

```mermaid
sequenceDiagram
    participant C as ScriptMapLibreGeoJson
    participant M as Map
    participant S as Style
    M->>S: setStyle() creates a new Style
    S-->>C: styledataloading (_loaded=false)
    Note over C: style is not mutable
    S->>S: _load() sets _loaded=true, adds style sources
    S-->>C: styledata
    S-->>C: style.load (isStyleLoaded=false)
    Note over C: addSource is legal here, so re-add now
    S-->>C: idle (isStyleLoaded=true)
    Note over C: safety net for a sync that arrived mid-swap
```

**Fix.** The component tracks style readiness from the map lifecycle instead of `isStyleLoaded()`:

- `style.load` and `load` mark the style mutable and run the sync.
- `styledataloading` marks it not mutable, because a full style reload starts there.
- `idle` retries a sync that was skipped mid-swap. MapLibre fires `idle` only when `Map.loaded()` is true, so the style is ready.

The `isStyleLoaded()` check is kept as an extra allow, never as the only gate. A skipped sync still clears the applied signature, so a later flip back to the last applied value rebuilds. That is the failure mode #901 had to repair once.

Both `setStyle` paths are covered. A full reload fires `styledataloading` then `style.load`. A diff swap keeps `_loaded` true and `Style.setState()` fires `style.load` at the end.

## PC-10: a paint property could not follow a theme

**Symptom.** A consumer could not drive a layer colour from a colour mode. Any content change to `layers` dropped the source and re-added it, which discarded the cluster index and flashed the map. The app therefore declared `transparent` as the layer colour and wrote every real colour through `setPaintProperty` on the raw map. That cost a whole module.

This is the deferred half of PC-1. #901 landed the structural comparison and left the layer diffing as follow-up.

**Fix.** The layer array is now split into two signatures:

- **Structure**: the source id, source options, `beforeId`, and every layer key except `paint`, `layout` and `filter`. A change rebuilds the source and layers, as before.
- **Style**: `paint`, `layout` and `filter`. A change alone routes to `setPaintProperty`, `setLayoutProperty` and `setFilter` on the layers already on the map. The source keeps its data and its cluster index.

Anything the component cannot apply in place, including a new layer id, type, order, `minzoom` or `source-layer`, falls into the structure signature and still rebuilds. A themed map is now a plain `computed`.

**Deferred.** A `sourceOptions` cluster change still rebuilds the source. Routing it to `source.setClusterOptions()` needs async handling and its own error path, so it is follow-up work.

**Hover cursor.** An in-place update drops the hover cursor, exactly as a prop-driven rebuild does on this base. New paint, layout or filter values can empty the spot under a stationary pointer, and MapLibre only re-evaluates on the next pointer move. The component never invents a `mouseleave`.

## Observable contract change

A failed in-place `paint`, `layout` or `filter` update emits `error` and **leaves** the source and layers on the map. A failed rebuild still removes them. The failed values are not recorded, so a corrected layer array applies again without a remount. A test covers both.

This also corrects a stale `error` contract doc left over from #901, which said only that the component removes its own source and layers. That was true then and is incomplete now. The JSDoc describes both outcomes.

## Tests

`test/nuxt-runtime/maplibre-components.nuxt.test.ts`:

- re-adds the source when `style.load` fires before the style reports loaded
- retries a sync skipped by an unloaded style once the map goes idle
- updates paint, layout and filter in place when the layer ids and types hold
- keeps the source on the map when an in-place layer update fails
- rebuilds the source when the layer structure changes

The PC-9 tests were written first and failed against #903's head.

`pnpm lint`, `pnpm typecheck` and the full suite pass locally, including the `typecheck` vitest project.


harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-16 08:13 UTC.

- A failed rebuild that also changes paint/layout/filter emits two identical error events for one prop change, contradicting the documented one-error contract. Next: In the style watcher fallback, skip the second trySyncResources when the structure watcher already attempted the same rebuild, for example by recording the attempted structure signature before syncing.
- The idle safety-net rebuild drops the hover cursor that the equivalent style.load rebuild keeps. Next: Call trySyncResources(map, true) from onIdle so the equality guard in bindLayerEvents decides whether to carry the hover.

547844c6-76fe-4bf1-8639-943affba52be

PR #909, opened 2026-09-16
fix(maplibre): emit invalid layer values and fix canvas a11y

harlan-zw:
### 📚 Description

Not breaking. No prop, emit or type is removed.

The `error` emit on `ScriptMapLibreGeoJson` from #901 shipped in `2.0.0-beta.7`. It does not work in a real browser for the most common failure. I caught it running beta.7 in a real Nuxt app on MapLibre 6.10.0. With `'circle-color': ['not-an-operator', ...]` on a layer, the app got no `error` emit, no console output, and a map with the `sites` layer missing. That is the silent blank map #901 set out to fix.

The mocked tests passed because a mocked `addLayer` always succeeds. MapLibre never throws for an invalid paint, layout or filter value. `addLayer`, `setPaintProperty`, `setLayoutProperty` and `setFilter` fire a map `error` event and skip the change.

**Invalid layer values (PC-12).** An error MapLibre fires during the component's own style call now fails that call, so the existing rollback and emit run. A later map error is emitted only when its `sourceId` or `layer.id` belongs to the component. Basemap, tile and raw-map errors stay on the map's `error` emit.

I did not match the `layers.<id>.` message prefix, which is what the app does today. `addLayer` validation errors carry nothing else, and the prefix is ambiguous once a layer ID contains a dot: `layers.a.` also matches layer `a.b`.

**Canvas accessibility (PC-14).** MapLibre gives its canvas `tabindex="0"`, `role="region"` and `aria-label="Map"`, nested inside the component's own labelled region.

- The canvas always loses `role="region"`, so each map has one landmark.
- MapLibre's keyboard handler pans and zooms only while the canvas has focus. While that handler is enabled, the canvas stays a tab stop.
- With `options: { keyboard: false }` or `:interactive="false"`, the canvas leaves the tab order and the accessibility tree.

Open question: the focused canvas keeps MapLibre's own `aria-label="Map"` with no role. I kept it because it follows MapLibre's locale option. It could take the component's `aria-label` instead.

**Real-browser harness.** `test/fixtures/maplibre` runs real MapLibre in Playwright Chromium. Styles are inline objects with only a `background` layer, so nothing waits on the network. One test records the WebGL renderer as a test annotation, so CI logs show which GL implementation ran. Locally that is SwiftShader.

`registry-types.json` is regenerated. That also fixes drift from #907, which changed the `layers` description without regenerating the committed file.

Once released, the app can delete `isSiteLayerError`, its test, and the `@error` filter on the map. It can delete the canvas attribute stripping if it also passes `keyboard: false`.


harlan-zw:
The new real-browser tests fail against `origin/main` (`6fd046e6`), before the fix:

```
 × maplibre in a real browser > emits error for an invalid paint value on first add, and only from the owning component
   → expected 0 to be greater than 0
 × maplibre in a real browser > emits error for an invalid paint value set on a live layer
   → expected [] to deeply equal [ StringMatching{…} ]
 × maplibre in a real browser > keeps a keyboard-operable canvas in the tab order without a nested landmark
   → expected { regions: 2, role: 'region', …(3) } to match object { regions: 1, role: null, …(1) }

 Test Files  1 failed (1)
      Tests  3 failed | 1 passed (4)
```

With the fix, all 4 pass. Local renderer annotation: `ANGLE (Google, Vulkan 1.3.0 (SwiftShader Device (Subzero) (0x0000C0DE)), SwiftShader driver)`.

harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-16 10:19 UTC.

The pull request closed.

PR #910, opened 2026-09-16
feat(maplibre): add scale, geolocate, fullscreen, attribution controls

harlan-zw:
### 🔗 Linked issue

Papercut PC-15 (first noted in PC-3).

### ❓ Type of change

### 📚 Description

Only `ScriptMapLibreNavigationControl` had a component. Every other standard control needed `@ready`, a manual `addControl`, and a manual `removeControl`.

This adds four components that follow `ScriptMapLibreNavigationControl`: a `position` prop, an `options` prop typed from the MapLibre v6 constructor, add on map ready, remove on unmount.

- `<ScriptMapLibreScaleControl>`
- `<ScriptMapLibreGeolocateControl>`
- `<ScriptMapLibreFullscreenControl>`
- `<ScriptMapLibreAttributionControl>`

Architecture: four new control components on the existing useMapLibreResource lifecycle, with public types from @nuxt/scripts

#### Option changes

The components read `options` and `position` once, like `NavigationControl`. To apply a change, change the component `key`. One exception follows the Marker and Popup rule for public setters: `ScaleControl` forwards `options.unit` through `setUnit()`.

#### GeolocateControl

- Re-emits all seven control events with MapLibre's payloads: `geolocate`, `error`, `outofmaxbounds`, `trackuserlocationstart`, `trackuserlocationend`, `userlocationfocus`, `userlocationlostfocus`.
- A denied prompt already fires `error` with `code` 1.
- A permission that was denied **before** the control loads fires nothing in MapLibre. It only disables the button and logs a warning. The component runs the same check and emits `unavailable` with `'permission-denied'` or `'unsupported'`.

#### AttributionControl

MapLibre adds a default attribution control to every map. A second control would duplicate it. Removing it would strip required OpenFreeMap and OpenStreetMap credits.

- On mount, the component removes the map's own `AttributionControl` and adds its own.
- On unmount, it adds the map's control back. Attribution shows exactly once in both states.
- If adding its own control throws, it restores the default before the error propagates.
- If the map was already removed (`map._removed`), it restores nothing. If consumer code removed the component's control, it still restores the default.
- The component `options` override the default control's options. Credits from every replaced control, including `attributionControl.customAttribution`, stay unless the component sets its own.
- `options` cannot remove source attribution from the style.

It finds the default control through `map._controls`, which is typed in `maplibre-gl.d.ts` and is the list `hasControl()` reads.

Data flow: the attribution component removes the default control on mount and restores it on unmount

#### Hydration fix for all five controls, including NavigationControl

Every control template held only an HTML comment. A production build strips template comments. The server then rendered nothing, but the client render returned `null`, which expects a comment node. Hydration logged `Hydration completed but contains mismatches.` `nuxt dev` keeps comments, so dev mode never showed it.

**Chosen shape:** no `<template>`, and `defineOptions({ render: () => null })`.

- The client normalizes `null` to a comment vnode.
- The component has no `ssrRender`, so the server renderer calls `render()`, normalizes `null` the same way, and writes `<!---->`.
- It does not depend on the compiler `comments` option, and it states "renders nothing" in code.
- `<div v-if="false" />` also matches on both sides, but it reads as a hack.
- An empty `<template>` has the same bug as a comment-only one, in dev and production.

Proof on a real `nuxt build` of `test/fixtures/maplibre`:

- Before: the slot rendered `<!--[--><!--]-->` for five controls. The client logged `error: Hydration completed but contains mismatches.`
- After: the slot renders `<!--[--><!----><!----><!----><!----><!----><!--]-->`. The client bundle has `render:()=>null`. The console is empty.

**Category guard:** `test/e2e/maplibre-hydration.test.ts`.

- `@nuxt/test-utils` builds inside the Vitest worker. There, `@vue/compiler-core` loads its development build, which keeps comments, so a plain e2e build cannot catch this bug.
- The test sets `vue.compilerOptions.comments: false`. A probe assertion proves a template comment is stripped from the server HTML.
- Reverting `ScriptMapLibreScaleControl` to the comment-only template fails the test. Restoring it passes.

Other components with the same shape are out of scope for this PR and should adopt the same fix.

#### Types

`ScriptMapLibre{Scale,Geolocate,Fullscreen,Attribution}ControlProps` and the Geolocate and Fullscreen emit types are exported from `@nuxt/scripts` and `#nuxt-scripts/types`.

### ✅ Verification

- Lifecycle tests in `test/nuxt-runtime/maplibre-controls.nuxt.test.ts`. They use a map mock with MapLibre's real control list semantics.
- `test/types/maplibre-controls.test-d.ts` imports only from `@nuxt/scripts`. With `ScriptMapLibreScaleControlProps` removed from the export, it fails.
- Real Chromium on the new `playground/pages/third-parties/maplibre-controls.vue` with OpenFreeMap:
  - Each control renders in its configured corner and is gone after unmount.
  - Attribution count is 1 with the component, 1 after unmount (default back in bottom-right), and 1 after remount.
  - The attribution text includes OpenFreeMap and OpenStreetMap in every state.
  - `unit` switches the scale from `1 km` to `1 mi`.
  - A denied prompt emits `error` with code 1. A pre-denied permission emits `unavailable: permission-denied`. A granted permission emits `geolocate`.
  - Fullscreen emits `fullscreenstart` and `fullscreenend`.

Non-breaking: all changes are new components, new types, and docs.


harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-16 10:58 UTC.

The pull request closed.

PR #911, opened 2026-09-16
feat(maplibre): add mousemove, dblclick and sourceready emits

harlan-zw:
Stacked on #909.

### 📚 Description

Not breaking. Three emits are added to `ScriptMapLibreGeoJson`: `mousemove`, `dblclick` and `sourceready`.

The group hover from #903 shipped in `2.0.0-beta.7` and **does not work in a real browser**. I caught it running beta.7 in a real Nuxt app on MapLibre 6.10.0. The component binds `mouseenter` and `mouseleave` across its whole layer array, and MapLibre fires those only at the edge of the group. With a site touching a cluster in another layer:

```
pointer onto site                 hovered=[2817331025242450]  cursor=pointer
slid straight onto the cluster    hovered=[2817331025242450]  stale, the site stays lit
cluster -> cluster -> site        hovered=[]                  stale, the site is not lit
off both                          hovered=[]                  cursor=""
```

The mocked tests called the stored listener directly, so they could not see MapLibre's group semantics. The `cursor` prop was fine.

**`mousemove` (PC-11).** It fires on every move over any owned feature, with `features` topmost first. With `mouseleave`, that is enough to track the feature under the pointer. I picked it over per-layer `mouseenter`/`mouseleave` because per-layer edges still miss a move between two touching features in the **same** layer. That is the site-to-site case in the app. `mouseenter` and `mouseleave` stay, and their docs now say they are group edges.

**`dblclick`.** The emit is synchronous, so `event.preventDefault()` still stops MapLibre's double-click zoom. The app found that an `easeTo` started inside a `dblclick` listener is cancelled. The cause: after dispatching the event, MapLibre's `HandlerManager.handleEvent` calls `camera.stop(true)` whenever a handler is active. Calling `preventDefault()` marks the map event handler active, so it happens either way. A camera move started on the next tick survives. I checked both in Chromium: zoom stays at 4 with a synchronous `easeTo`, and reaches 6 with `nextTick`.

I documented that rather than deferring the emit, because a deferred emit would break `preventDefault()`. A `doubleClickZoom` prop scoped to the component's layers is possible. It would duplicate `preventDefault()`, so I left it out. Happy to add it if you prefer a declarative switch.

**`sourceready` (PC-13).** The map's `styleload` fires before the component re-adds its source, so feature state restored there targets a missing source. The app works around it with a `nextTick`. `sourceready` fires after every successful add: the first add, a style swap and a prop rebuild. Its payload is `{ map, sourceId }`. The emit runs after the style change, so a map error from the handler is never attributed to the component.

The real-browser fixture from #909 now also covers hover across touching features, `dblclick` with a replaced zoom, and `sourceready` after a style swap. The swap runs both ways MapLibre applies a style: as a diff and as a full reload.

`registry-types.json` is regenerated for the new emits.

Once released, the app can delete:

- the raw `mousemove`, `mouseleave` and `dblclick` listeners in `_MapPanel.vue` `onReady`. The `dblclick` handler must still defer its `easeTo`.
- the `nextTick` in `onStyleLoad`, by moving `syncFeatureStates` to `@sourceready`.


harlan-zw:
The new real-browser tests fail on the parent branch (`340eafd0`), before these emits:

```
 FAIL  maplibre in a real browser > follows the pointer between touching features in different layers
AssertionError: expected 'none' to be 'west' // Object.is equality
 FAIL  maplibre in a real browser > emits dblclick and lets the consumer replace the default zoom
AssertionError: expected '{"mouseenter":1,"mouseleave":0,"dblcl…' to contain '"dblclick":1'
 FAIL  maplibre in a real browser > emits sourceready after a diff style swap so feature state can be restored
 FAIL  maplibre in a real browser > emits sourceready after a full style swap so feature state can be restored
AssertionError: expected 0 to be greater than 0
      Tests  4 failed | 4 passed (8)
```

With the emits: `Tests  8 passed (8)`.

I also checked the `dblclick` camera claim by hand in the same fixture. After `preventDefault()`, a synchronous `easeTo({ zoom: 6 })` left the map at `{"zoom":4,"center":[0,0]}`. The same call in `nextTick` ended at `{"zoom":6,"center":[-1.2003,0]}`.

PR #914, opened 2026-09-16
fix(maps): render renderless map components as null

harlan-zw:
### 🔗 Linked issue

Papercut PC-18. Follow-up to #910, which fixed the same bug in the five MapLibre controls.

### ❓ Type of change

#### ⚠️ Google Maps mismatches in `nuxt dev` too

`ScriptGoogleMapsPolygon`, `Polyline`, `Circle`, `Rectangle` and `GeoJson` had an empty `<template>`. An empty template compiles to an empty `ssrRender` on the server and `return null` on the client, even in development.

Measured with `@nuxt/test-utils` in `dev: true` mode, before this fix:

```
[Vue warn]: Hydration children mismatch ... Server rendered element contains fewer child nodes than client vdom.
  at <ScriptGoogleMaps ...>
Hydration completed but contains mismatches.
```

Under `nuxt dev`, the MapLibre and Leaflet pages logged no hydration message. After this fix, no page logs one.

#### Production builds: all eight components

`ScriptMapLibreGeoJson`, `ScriptLeafletGeoJson` and `ScriptLeafletTileLayer` had a template with only a comment. A production build strips template comments, so the server renders nothing. The client expects a comment node, and hydration reports a mismatch.

#### Fix

Each component now has no `<template>` and uses `defineOptions({ render: () => null })`. This is the shape #910 proved for the MapLibre controls.

- The client normalizes `null` to a comment vnode.
- With no `ssrRender`, the server calls `render()`, normalizes `null` the same way, and writes `<!---->`.
- It does not depend on the compiler `comments` option.

#### Test infrastructure finding

`@nuxt/test-utils` builds inside the Vitest worker with `NODE_ENV=test`. The Vue compiler then loads its development build and keeps template comments. This repo's e2e tests therefore could never catch a production-only hydration bug.

The new `test/e2e/map-hydration.test.ts` sets `vue.compilerOptions.comments: false`. A probe test proves that each page's template comment is stripped from the server HTML. Without the probe, a silent fallback to development behaviour would make the hydration test pass for the wrong reason.

The fixture `test/fixtures/map-hydration` loads no map SDK, tiles, glyphs or sprites. Hydration compares the server HTML with the client's first render, which happens before any script loads. The Google Maps page uses a placeholder string for `api-key`. No real key is involved.

### ✅ Verification

Production-mode hydration test, per library:

| Page | Before fix | After fix | One component reverted |
|---|---|---|---|
| `/maplibre` (`GeoJson`) | fail: `Hydration completed but contains mismatches.` | pass | `ScriptMapLibreGeoJson`: fail |
| `/leaflet` (`GeoJson`, `TileLayer`) | fail: same message | pass | `ScriptLeafletTileLayer`: fail |
| `/google-maps` (5 components) | fail: same message | pass | `ScriptGoogleMapsPolygon`: fail |

The Google Maps mismatch reproduces without an API key.

- `pnpm lint`: 0 errors. `pnpm typecheck`: exit 0.
- Full suite: `Tests 1206 passed | 22 skipped | 3 todo`, `Type Errors no errors`. Two e2e files failed their fixture builds on CDN fetches (`analytics.ahrefs.com`, `snap.licdn.com`). A rerun of those two, with all map e2e files, passed: `Test Files 5 passed`, `Tests 27 passed`.

Non-breaking.


harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-16 12:24 UTC.

8312c038-94f6-4183-b421-0c82bcc5af36

PR #913, opened 2026-09-16
fix(maplibre): keep credits from every replaced attribution control

harlan-zw:
### 🔗 Linked issue

Follow-up to #910. This addresses the CodeRabbit finding posted on #910 two seconds after it merged. `2.0.0-beta.8` shipped without this fix.

### ❓ Type of change

- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

`<ScriptMapLibreAttributionControl>` replaces every `AttributionControl` on the map. It copied `customAttribution` only from the first replaced control.

If a map had a second attribution control with its own credits, those credits disappeared while the component was mounted. Unmounting restored them.

The component now merges the `customAttribution` values of all replaced controls, without duplicates. A `customAttribution` set on the component still replaces them.

A single inherited string now reaches MapLibre as a one-item array. MapLibre renders both forms the same way.

### ✅ Verification

- A new test in `test/nuxt-runtime/maplibre-controls.nuxt.test.ts` failed before the fix (`1 failed | 14 passed`) and passes after it (`15 passed`).
- `pnpm lint`: 0 errors. `pnpm typecheck`: exit 0.
- Full suite: `Test Files 118 passed`, `Tests 1217 passed | 6 skipped | 3 todo`, `Type Errors no errors`.
- `test/e2e/maplibre.test.ts` and `test/e2e/maplibre-hydration.test.ts`: `Tests 10 passed (10)`.

Non-breaking.

PR #916, opened 2026-09-16
fix(maplibre): replace default attribution through public API

harlan-zw:
### 🔗 Linked issue

Follow-up to #910 and #913.

### 📚 Description

`<ScriptMapLibreAttributionControl>` found the map's default attribution control by reading `map._controls`. The field shows up in MapLibre's `.d.ts`, but nothing promises it stays. If a release renames it, attribution shows twice or fails to come back on unmount, and OpenFreeMap and OpenStreetMap require it to be visible.

MapLibre has no public way to get the default control: the constructor creates it and keeps no public reference. So `<ScriptMapLibreMap>` now creates that control itself, with the same options, and shares it through the map context. The attribution component swaps it with `hasControl`, `removeControl` and `addControl`. It also listens for the `remove` event, so it no longer reads `_removed`.

Behaviour change: the component replaces only the map default. An attribution control added by your own code with `map.addControl()` now stays, so attribution can show twice, but it never disappears. The docs say so.

Open question: if `maplibreLogo` is on and `logoPosition` is `bottom-right`, the logo and the default attribution swap vertical order in that corner, because the default control is added after the constructor. I left it. Say if it matters.


harlan-zw:
Checked by hand, since CI cannot show it:
- I disabled the restore call in the attribution component and rebuilt. The new real-browser test failed with `expected [] to deeply equal [ { …(2) } ]`, so it catches lost attribution.
- The updated runtime tests hide the fake map's control list. On `main` they fail with `TypeError: Cannot read properties of undefined (reading 'filter')`, which is the renamed `_controls` case.

harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-21 15:13 UTC.

ad196754-b1c3-4440-a5a3-d004893debb9

PR #917, opened 2026-09-16
feat(maplibre): add bounds prop and update cluster options in place

harlan-zw:
### 📚 Description

Two MapLibre papercuts from building a real map app (PC-16, PC-17).

**`bounds` prop.** `center` was required, so framing the data meant computing a centre I never used, then calling `fitBounds` from `ready`. Two sources of truth for the first camera. `ScriptMapLibreMap` now takes `bounds` and `fitBoundsOptions`.

- Types: `ScriptMapLibreMapProps` is shared props plus a union of a center camera and a bounds camera. A map with neither is a type error. Vue merges the union at runtime, so both compile to `required: false` and I skipped a dev warning.
- Precedence: with both set, `bounds` sets the initial center and zoom, same as MapLibre's own `bounds` option. The fit keeps the `bearing` prop, because MapLibre's fit resets it to 0.
- Later changes: new coordinates refit without animation, like the other camera props. A new array with the same coordinates does nothing, so a parent render can't undo a user pan. `fitBoundsOptions` alone does not refit.
- The registry types generator now reads union and intersection props. Without that, the docs table lost every map prop.

**Cluster options in place.** A `clusterRadius` change removed and re-added the source, dropping the cluster index (deferred in #901 and #907). A change to only `cluster`, `clusterRadius` or `clusterMaxZoom` now goes to `setClusterOptions()`. Those are the only three MapLibre 6.9.1 accepts. `clusterMinPoints`, `clusterProperties`, and removing radius or max zoom still rebuild. A failure emits `error`, and a failure from a superseded call is ignored.

Architecture after this change: bounds reaching the Map constructor, the GeoJson component calling setClusterOptions, and the docs type generator

Not sure about: refitting on `bounds` change only when coordinates differ. It's the least surprising option I found, but a "zoom to data" button that re-sends the same bounds won't move the camera; that still needs the exposed `fitBounds`.

### ⚠️ Breaking Changes

None at runtime or in templates; every existing caller passes `center`. `ScriptMapLibreMapProps` is now a type alias, so `interface X extends ScriptMapLibreMapProps` would need an intersection. Nothing in this repo does that.


harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-17 03:31 UTC.

155174be-07d4-4a53-b860-20d895ce1e34

PR #918, opened 2026-09-16
test(e2e): build fixtures with template comments stripped

harlan-zw:
## Why

`@nuxt/test-utils` builds fixtures inside the Vitest worker, where `NODE_ENV` is `test`. `@vue/compiler-core` then loads its dev build, and its `comments` option defaults to `true`. A production build strips template comments. So no e2e fixture could fail on a production-only hydration bug, such as the comment-only templates that #910 and #914 fixed. Those PRs patched only their own two fixtures.

## What

- `test/utils/setup-fixture.ts`: `setupFixture()` wraps `setup()` and forces `vue.compilerOptions.comments: false` into `nuxtConfig`. Every test in `test/e2e` and `test/e2e-dev` now calls it.
- `eslint.config.mjs`: `no-restricted-imports` blocks `setup` from `@nuxt/test-utils/e2e` in `test/e2e/**` and `test/e2e-dev/**`. A new test cannot skip the wrapper by accident.
- `test/e2e/production-compile.test.ts` with fixture `test/fixtures/production-compile`: one shared probe. It fails if the server HTML keeps a template comment. It also fails if a comment-only template component stops reporting a hydration mismatch.
- `map-hydration` and `maplibre-hydration` lose their own override and probe tests. Their hydration tests stay.

Why a wrapper around `setup()`: every e2e test already passes `nuxtConfig` overrides through `setup()`, and no fixtures share a config layer. A layer would need an `extends` line in all 20 fixtures and would also change `nuxt dev` on those fixtures. The wrapper is one place, and the lint rule enforces it.

## Proof

- Probe with the override removed from `setupFixture()`: both probe tests fail (`expected '<!DOCTYPE html>…' not to contain 'production-compile-probe'`, `expected [] to not deeply equal []`). Restored: both pass.
- No fixture outside the two map fixtures renders one of the 13 fixed components. In `map-hydration`, a temporary revert of `ScriptLeafletTileLayer.vue` to a comment-only template fails `hydrates /leaflet without a mismatch` with `Hydration completed but contains mismatches.` The shared mechanism alone still catches it.
- Full e2e suite: 20 of 21 files pass. The failure is the known `reCAPTCHA` network flake; alone it passed once and failed once with this change.

## Findings

**No new comment-caused hydration mismatch.** A temporary sweep visited every page of every fixture with hydration details on. 8 pages report a mismatch, and all 8 report it identically with comments kept, so none comes from this change. They are text mismatches in fixture pages, not comment handling, so this PR does not fix them:

- `basic` `/reload-trigger`: `#use-call-count` is `0` on the server and `1` on the client. The `use()` callback increments a ref during client setup.
- `basic` `/tpc/vercel-analytics`, `/tpc/speedcurve-unregistered`, and `speedcurve` `/tpc/speedcurve*` (5 pages): `#status` is `awaitingLoad` on the server and `loading` on the client.

No test asserts on hydration for these pages today.

**Other production-only compile differences remain under `NODE_ENV=test`.** Vite derives `config.isProduction` from `process.env.NODE_ENV`, so `@vitejs/plugin-vue` compiles SFCs in dev mode. The same component built with `NODE_ENV=test` and `NODE_ENV=production` differs:

| | `NODE_ENV=test` | `NODE_ENV=production` |
|---|---|---|
| `<script setup>` template | separate render, `setup()` returns a bindings object (`__isScriptSetup`) | inlined render closure |
| `defineProps` runtime types | `{ type: String, required: false }` | `{}` |
| `__file` with absolute path | in client and server bundles | absent |

Nuxt defines `process.env.NODE_ENV` for the bundles from `vite.mode`, so the Vue runtime in the bundles is already the production build. This PR changes only comment handling. Setting `NODE_ENV=production` for the build would close all three gaps, but it changes more than comments, so it is left for a follow-up.


harlan-zw:
### 🤖 MERGED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. It is not Harlan's personal review or approval. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). Last updated: 2026-09-21 13:46 UTC.

83661ade-28da-4875-9682-41aa2ac8dd71

PR #923, opened 2026-09-21
chore(pulse-analytics): use an inline SVG logo instead of a PNG URL

uz1mani:
The `pulseAnalytics` entry shipped a URL to a 64px PNG. Swapped for the vector mark, inline like the other 37 logos.

One file, one line: `packages/script/src/registry-logos.ts`.

## Why

- It is one of only three URL-based logos in the registry, and the only raster one (`clarity` and `tawkTo` are the other two URLs).
- The sidebar icon renders at 18px and cards go up to 64px, so a 64px PNG is upscaled on any retina display. It reads soft next to 37 vector neighbours.

## Size

1,369 bytes — rank 10 of 38 inline logos, against a median of 805.

The source was an auto-trace at 5,839 bytes, which would have been rank 2. `svgo --precision=1` cut it 77%; a pixel diff of before against after shows differences confined to edge antialiasing, with no geometry moved.

The mark also gained a `viewBox` it did not have — the source declared `width`/`height` only, so it could not scale inline. Measured ink bounds were `x 32.8..163.5, y -0.7..192.6`; the viewBox is that squared.

## Test

`test/unit/registry-logos.test.ts` — 2 passed, no type errors.

Control: replacing the value with a `data:image/png;base64,…` string fails it with `pulseAnalytics logo should be SVG or image URL`, so the pass is meaningful.


harlan-zw:
### 🤖 REVIEW PAUSED

This pull request is from an outside contributor. Add the `harlan-agent-review` label to approve automated review and verified repairs for head commit `6d8cd377813e`.

uz1mani:
Worth linking: this logo will not be visible on scripts.nuxt.com until the site picks up a build past `2.0.0-beta.6`, which is what #924 is about.

The same staleness is why `pulse-analytics` currently renders in the Utility group with a placeholder icon rather than under Analytics — the entry on `main` sets `'analytics'`, but the site is reading registry data from beta.6, which predates it.

So this PR is correct on its own terms and testable via the unit test, but the visible result depends on #924.

harlan-zw:
Thanks :)

Issue #720, opened 2026-04-20
FathomAnalytics: visitors identified as bots since v1


IsraelOrtuno:
### 🐛 The bug

Since v1, FathomAnalytics seems to identify visitors as bots. Raising this up just in case someone else is facing this problem or can add more info. I downgraded to 0.x for now.

<img width="225" height="72" alt="Image" src="https://github.com/user-attachments/assets/3ba9f694-6f0f-4a24-b17a-9a009008ba1c" />

Simplest setup:

```
registry: {
  fathomAnalytics: {
    site: 'xxx',
    trigger: 'onNuxtReady'
  }
}
```

Digging more into it, will provide more info when I have.

With v0.x
<img width="1505" height="142" alt="Image" src="https://github.com/user-attachments/assets/f46a2019-a429-4606-ac7c-81d01e35a8b1" />

With v1.0+
<img width="1507" height="146" alt="Image" src="https://github.com/user-attachments/assets/6ea05a8d-57e0-4fb7-ac62-988ff8415e77" />

It's very likely this could be issue from Fathom side and how they detect bots.

### 🛠️ To reproduce

..

### 🌈 Expected behavior

Track visitors correctly.

PR #927, opened 2026-09-23
feat: Statable Analytics

key-arg:
### 📚 Description

[Statable](https://statable.com/) is privacy-first web analytics by Key Arg B.V., a Dutch company. This adds it to the registry as `statableAnalytics` / `useScriptStatableAnalytics()`, with `siteId`, `host`, `trackingApi` and `props` mapped onto the tracker's script URL and `data-*` attributes, and the tracker's single `t(name, props?)` call exposed through the composable's proxy.

Capabilities are `bundle: true` and nothing else, the same family as Fathom (#720) and Pulse (#899). Statable works out visitor identity on its server from a keyed hash of the connecting IP and user agent, so beacons routed through the Nuxt server would all arrive from one IP and collapse every visitor into a single identity.

One thing worth knowing about bundling: the tracker reads the site id from its own `src` path (`/js/{siteId}/s.js`) and posts to the origin it was served from. A bundled copy is served from the Nuxt origin, so the composable pins `data-id` and `data-tracking-api` on the script element. No SDK patch is needed.

The tracker keeps no identifier in the browser. The only thing it touches in `localStorage` is an opt-out flag it reads, and it honours Do Not Track and Global Privacy Control on its own. Measured with `scripts/generate-sizes.ts`: 3 KB over the wire, one request, no cookies.

Playground and fixture use `siteId: '123456'`, which is not a registered site, so the API rejects the events. A real id here would record playground visits.

Tracker reference: https://statable.com/docs/developers/tracking-script/

Disclosure: I work on Statable.


harlan-zw:
### 🤖 QUEUED

> [Harlan Agent Kit](https://github.com/harlan-zw/harlan-agent-kit) posted this automated review. [AI open source policy](https://harlanzw.com/blog/ai-in-open-source). The Dashboard shows the exact Queue position.

Next: Review starts when an Agent is free.

